PR #8, opened 2019-08-19
Bug 1741822: bypass kube authn/z on always allowed paths

stlaz:
We are using kube authnentication to wrap the oauth-server
endpoints and, typically, let the oauth-server itself fetch
the clientCA from the cluster for cluster admin
authenticatication.

When using Request Header identity provider, we also need to use
client cert authentication. However, oauth-server's own authentication
is logically preceded by the kube one. Since this one is configured
to fail on errors, and the client CA for Request Header IdPs is
usually different from the one that issued cluster admin cert,
the kube cert authnetication fails sooner than the request is passed
to the oauth-server.

The behavior is fixed here by using a union authenticator of the
kube authenticator and an authenticator that will ignore kube
authentication errors on paths that any user is authorized to access
(as configured for the delegated authorization).

cc @enj @deads2k


stlaz:
`verify-deps` was failing on unknown gssapi repo dependency. While that moved to `openshift/gssapi`, I did not find why it should be included here (unless it appears in a very well-hidden build flag), so I just removed it.

pkg/cmd/oauth-server/server.go
```diff
@@ -2,13 +2,19 @@ package oauth_server
...
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authentication/authenticator"
+	"k8s.io/apiserver/pkg/authentication/request/anonymous"
+	unionauth "k8s.io/apiserver/pkg/authentication/request/union"
 	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	pathauthorizer "k8s.io/apiserver/pkg/authorization/path"
```
enj: nit: remove unnecessary alias

pkg/cmd/oauth-server/server.go
```diff
@@ -2,13 +2,19 @@ package oauth_server
...
+	"net/http"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authentication/authenticator"
+	"k8s.io/apiserver/pkg/authentication/request/anonymous"
+	unionauth "k8s.io/apiserver/pkg/authentication/request/union"
```
enj: nit: remove unnecessary alias

enj:
Oh and I want the e2e to prove this works :smiley:

PR #10, opened 2019-08-29
migrate to go module

sanchezl:
1. add dependency magnet for alpha-build-machinery
2. ran `go mod init github.com/openshift/oauth-server`
3. ran `go get k8s.io/apiserver@kubernetes-1.14.0`
4. ran `go mod edit -replace k8s.io/apiserver=github.com/openshift/kubernetes-apiserver@oauth-server-4.2-kubernetes-1.14.0`
5. ran `go mod tidy`
6. ran `go get` to get correct versions of openshift osin & osincli
7. ran `go mod edit -replace` to point `github.com/RangelReale/osin => github.com/openshift/osin`
8. ran `go mod edit -replace` to point `github.com/RangelReale/osincli => github.com/openshift/osincli`
9. removed github.com/openshift/osin* from `go.mod`
A. ran `go mod vendor`


deads2k:
All that go mod pinning looks questionable.  I think you should consider scrubbing that list down so we can match kube.

/lgtm

we'll start here.

deads2k:
verify-deps won't ever work.

PR #11, opened 2019-09-10
Bug 1750786: Fix bootstrap IdP challenges ruining RequestHeaders IdP

stlaz:
If RequestHeaders are configured for challenges, don't set up
the bootstrap IdP for challenges either. This would otherwise
add the WWW-Authenticate header on top of the redirecting request
which then in turn causes oauth-server internal errors.

### DIFF
--- a/pkg/oauthserver/oauth_apiserver.go
+++ b/pkg/oauthserver/oauth_apiserver.go
@@ -104,8 +104,10 @@ func NewOAuthServerConfig(oauthConfig osinv1.OAuthConfig, userClientConfig *rest
 		oauthConfig.IdentityProviders = append(
 			[]osinv1.IdentityProvider{
 				{
-					Name:            bootstrap.BootstrapUser, // will never conflict with other IDPs due to the :
-					UseAsChallenger: true,
+					Name: bootstrap.BootstrapUser, // will never conflict with other IDPs due to the :
+					// don't set it up as challenger if RequestHeaders IdP already is set that way
+					// this would set challenging headers and break RequestHeaders IdP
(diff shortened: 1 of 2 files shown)


pkg/oauthserver/oauth_apiserver.go
```diff
@@ -211,6 +213,15 @@ func isHTTPS(u string) bool {
 	return err == nil && parsedURL.Scheme == "https"
 }
 
+func isRequestHeadersSetAsChallenger(providers []osinv1.IdentityProvider) bool {
+	for _, p := range providers {
+		if _, isRequestHeaders := p.Provider.Object.(*osinv1.RequestHeaderIdentityProvider); isRequestHeaders && p.UseAsChallenger {
```
enj: nit `isRequestHeader`

pkg/oauthserver/oauth_apiserver.go
```diff
@@ -211,6 +213,15 @@ func isHTTPS(u string) bool {
 	return err == nil && parsedURL.Scheme == "https"
 }
 
+func isRequestHeadersSetAsChallenger(providers []osinv1.IdentityProvider) bool {
```
enj: nit: `isRequestHeaderSetAsChallenger`

stlaz:
Fixed the nits and build. This scenario should also be included in https://github.com/openshift/origin/pull/23662 once it's done

enj:
@stlaz I opened https://bugzilla.redhat.com/show_bug.cgi?id=1750809 for 4.1.z backport.

PR #14, opened 2019-09-24
move to kube-1.16

sanchezl:


go.mod
```diff
@@ -4,90 +4,52 @@ go 1.12
...
-	github.com/modern-go/reflect2 v1.0.1 // indirect
-	github.com/munnerz/goautoneg v0.0.0-20120707110453-a547fc61f48d // indirect
-	github.com/natefinch/lumberjack v2.0.0+incompatible // indirect
 	github.com/onsi/ginkgo v1.9.0 // indirect
 	github.com/onsi/gomega v1.7.0 // indirect
-	github.com/openshift/api v0.0.0-20190814194116-a94e914914f4
-	github.com/openshift/client-go v0.0.0-20190813201236-5a5508328169
-	github.com/openshift/library-go v0.0.0-20190829104452-2cb04072d0f5
-	github.com/pborman/uuid v0.0.0-20150603214016-ca53cad383ca // indirect
+	github.com/openshift/api v3.9.1-0.20190923092516-169848dd8137+incompatible
```
deads2k: v3.9.1 something looks very suspicious.

sanchezl: It's odd, but it does point to the correct commit.

sanchezl: I think it's just `go mod` being clever. From: https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning:

>  if a source code repository has a v2.0.0 or later tag for a file tree with no go.mod, the version is considered to be part of the v1 module's available versions and is given an +incompatible suffix when converted to a module version, as in v2.0.0+incompatible. The +incompatible tag is also applied to pseudo-versions derived from such versions, as in v2.0.1-0.yyyymmddhhmmss-abcdefabcdef+incompatible.

PR #19, opened 2019-10-23
Bug 1739262: RequestHeaders IdP: fix TLS handshake

stlaz:
OAuth server uses the APIServer wrapping for its endpoints. This
wrapping has its own RequestHeader IdP for the cluster admin login
with client cert. As a part of that, before the RequestHeader is
set, DelegatingAuthenticationOptions() adds its own ClientCA to
the serving info config.

The result of the above behavior is that during TLS handshake, when
the names for acceptable client certificate CAs are announced, the
CAs of the above mentioned apiserver RequestHeader IdP are announced.

When users configure their own RequestHeader IdP to be used in
OpenShift for the oauth-server to try to retrieve the users from,
oauth-server was not adding this RequestHeader's client CA name
among the acceptable client certificate CA names mentioned above.

This commit adds the client CA data from the user specified
RequestHeader IdP to 'GenericConfig.SecureServing.ClientCA',
which is later converted to tlsConfig object which defines the
data to be used in TLS handshakes.


pkg/oauthserver/auth.go
```diff
@@ -644,6 +644,12 @@ func (c *OAuthServerConfig) getAuthenticationRequestHandler() (authenticator.Req
 						return nil, fmt.Errorf("Error loading certs from %s: %v", provider.ClientCA, err)
 					}
 
+					// we need to add our CA data to secure serving as well to have the OAuth server
+					// advertise them for client auth during TLS handshake
+					if ok := c.GenericConfig.SecureServing.ClientCA.AppendCertsFromPEM(caData); !ok {
+						return nil, fmt.Errorf("Error adding certs from %s to secure serving: %v", provider.ClientCA, err)
```
mfojtik: lowecase error pls

deads2k:
This screams: "where is my e2e test".  As I recall, we lost this exact integration test when we went to a separate process in 4.1.  Can you add the e2e test we need?

stlaz:
I'm going to add an e2e test among the others in openshift/origin - https://github.com/openshift/origin/blob/master/test/extended/oauth/requestheaders.go, if that's ok. We could eventually bring all the oauth-server related tests here, but I would not like to do it with this PR.

I have a config for httpd container that I was using to reproduce the issue, I think with some modifications we should be able to use it for the test.

stlaz:
@deads2k I did not realize until now that the minimal e2e reproducer is actually quite simple - https://github.com/openshift/origin/pull/24053.

I can have the whole IdP fully set up for e2e a bit later if this reproducer is enough for now.

PR #23, opened 2019-11-19
Add bootstrap idp only when bootstrap user is enabled

vareti:

PR #26, opened 2019-12-09
Bug 1671604: Add bootstrap idp only when bootstrap user is enabled

stlaz:
adds bump + master rebase to #23


stlaz:
/retitle Bug 1671604: Add bootstrap idp only when bootstrap user is enabled

PR #30, opened 2020-01-27
Revert "Update login templates to PatternFly 4 design"

deads2k:
This reverts commit ff8ff138740ef1b2c485fd1b3d1e65f1eb60801e.

See if this puts serial green again


deads2k:
/hold

this is unlikely in comparison to the bound token change.

PR #31, opened 2020-01-30
Bug 1790998: switch version metric to component-base

sttts:


go.mod
```diff
@@ -3,50 +3,34 @@ module github.com/openshift/oauth-server
 go 1.12
```
stlaz: I think we want to switch this to `go 1.13`

tnozicka:
/hold
the tests are failing because it is being built with go1.12 instead of 1.13
```--> RUN make build --warn-undefined-variables
build flag -mod=vendor only valid when using modules

PR #32, opened 2020-02-11
move to openshift/build-machinery

stlaz:
/cc @marun @vareti


marun:
/lgtm

@stlaz Is this intended for 4.4 such that you're going to file the required bz? Or will it wait for 4.5?

stlaz:
looks like it's going to have to wait until 4.5 unless we get a BZ for the oauth-server :/

PR #34, opened 2020-02-25
Bug 1806620: reduce ldap connection timeout from 60s to 30s

vareti:
LDAP connection timeout was 60s. This might end up eating request timeout(60s) if there is an issue connecting to LDAP server causing the user to login failure without even trying other IDPs. As all the other IDP clients seems to be using 30s connection timeout, it is logical to set LDAP connection timeout to 30s too.

### DIFF
--- a/pkg/authenticator/password/ldappassword/ldap.go
+++ b/pkg/authenticator/password/ldappassword/ldap.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"runtime/debug"
+	"time"
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -42,6 +43,12 @@ type Authenticator struct {
 
 // New returns an authenticator which will validate usernames/passwords using LDAP.
 func New(providerName string, options Options, mapper authapi.UserIdentityMapper) (authenticator.Password, error) {
+	// BZ 1806620: Sometimes connecting to LDAP server might take long due to
+	// network issues causing the request to OAuth server to timeout(60s)
+	// before trying other IDPs even if the user can be authenticated using an
+	// IDP other than LDAP. Changing the global LDAP connection timeout here to
+	// 30s to fix this scenario and make it consistent with other OAuth IDPs
+	ldap.DefaultTimeout = 30 * time.Second
 	auth := &Authenticator{
 		providerName: providerName,
 		options:      options,


pkg/authenticator/password/ldappassword/ldap.go
```diff
@@ -42,6 +43,7 @@ type Authenticator struct {
 
 // New returns an authenticator which will validate usernames/passwords using LDAP.
 func New(providerName string, options Options, mapper authapi.UserIdentityMapper) (authenticator.Password, error) {
+	ldap.DefaultTimeout = 30 * time.Second
```
stlaz: Please add a comment about why you're doing this and that it is a global variable of the ldap module that we're writing into.

vareti: addressed it! please take a look!

sttts:
This needs a release note and documentation cc @bergerhoffer. 

@vareti can you link to a release note change? @bergerhoffer can give you pointers how to do that.

pkg/authenticator/password/ldappassword/ldap.go
sttts: So this means that out of 60 sec request timeout we have 30s left at worst? This still kills every request taking longer than 30s to be processed by etcd.

bergerhoffer:
@sttts @vareti Is this for 4.4 or 4.5? And are you considering this a bug fix, or more of a feature that you can fall back to a different IDP to log in (or was it only in this situation you couldn't fall back)?

pkg/authenticator/password/ldappassword/ldap.go
vareti: Yes, I agree. There are other failure cases too. If there are 3 IDPs configured(assuming it is a realistic scenario) and if the first two IDPs take 30s, request would be killed without trying the third IDP. 

I am not sure what repercussions increasing the request timeout would have. I didn't feel comfortable increasing it and so i ended up reducing the ldap timeout. Do you think it is better or there is a better way of solving it?

vareti:
@bergerhoffer This is a 4.4 bug fix. Fall back to different IDP is the expectation but the behavior seems to be different.

bergerhoffer:
Great thanks @vareti! I'll make a note of it in the release notes tracker just so we ensure it gets added, but what would help is if you could add a summary of the issue/resolution in the **Doc Text** field of the BZ. Thanks!

bergerhoffer:
@vareti Also looking at the BZ, it looks like the target release is 4.5. Should it be 4.4 if it's a bug fix for 4.4? Just wanting to make sure it shows up when we do our query for all 4.4 bug fixes.

sttts:
/lgtm
/approve

This is a little improvement. Not a real solution. At least the cluster does not break down completely.

PR #47, opened 2020-06-02
cleanup dead code in default auth handlers

vareti:
Cleaning up a conditional statement as it looks like a dead code and the boolean it is testing is evaluated much before [here](https://github.com/vareti/oauth-server/blob/485191656294201658b8014886683fd491911bc3/pkg/oauth/handlers/default_auth_handler.go#L83)

### DIFF
--- a/pkg/oauth/handlers/default_auth_handler.go
+++ b/pkg/oauth/handlers/default_auth_handler.go
@@ -115,11 +115,7 @@ func (authHandler *unionAuthenticationHandler) AuthenticationNeeded(apiClient au
 				w.WriteHeader(http.StatusFound)
 			default:
 				w.WriteHeader(http.StatusUnauthorized)
-				ctx := req.Context()
-				if !ok {
-					return false, fmt.Errorf("no context found for request to audit")
-				}
-				ev := request.AuditEventFrom(ctx)
+				ev := request.AuditEventFrom(req.Context())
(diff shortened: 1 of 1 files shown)


pkg/oauth/handlers/default_auth_handler.go
```diff
@@ -115,11 +115,7 @@ func (authHandler *unionAuthenticationHandler) AuthenticationNeeded(apiClient au
 				w.WriteHeader(http.StatusFound)
 			default:
 				w.WriteHeader(http.StatusUnauthorized)
-				ctx := req.Context()
-				if !ok {
```
stlaz, line 119: I'm some 90% sure this should've been `ctx == nil` originally

vareti: Yes. It does look suspicious. But the functional call `req.Context()` seem to suggest otherwise. How do you think we should proceed on this then?

stlaz:
For the record, the change dates back to https://github.com/openshift/oauth-server/commit/8f1a7ff451bba5a7728ef2d93309604267f09b13

@vareti (nit warning): please change the commit message from `cleanup dead code` to `clean dead code in default auth handlers`, I'll tag the PR afterwards

vareti:
@stlaz thanks for the review! updated the commit message. Please take a look.

PR #49, opened 2020-06-25
Use new access token inactivity timeout field

vareti:
This PR aims to use consistent field name for access token inactivity timeout. It uses the new fields that were added to token config  oauth cluster config and osin config.


stlaz:
/hold
contains fake bump

sttts:
"Pending — Not mergeable." What does this mean?

PR #52, opened 2020-09-02
Don't use kubeadmin as challenger if it's the only challenging IdP

stlaz:
Oftentimes the users configure IdPs such as OIDC that only allow
web-based flows. In that case it only makes sense that CLI logins
with `oc` redirect the users to the web flow when kubeadmin is the
only IdP that would otherwise make the oauth-server issue HTTP
basic auth challenge.


stlaz:
/close
this was marked as a too big of a behavior change

PR #55, opened 2020-09-04
Introduce a bootstrap realm for basic-challenge so it's unique.

sallyom:
/assign @stlaz 

for this: https://github.com/openshift/oc/pull/553

### DIFF
--- a/pkg/oauthserver/auth.go
+++ b/pkg/oauthserver/auth.go
@@ -355,8 +355,13 @@ func (c *OAuthServerConfig) getAuthenticationHandler(mux oauthserver.Mux, errorH
 				login.Install(mux, loginPath)
 			}
 			if identityProvider.UseAsChallenger {
+				realm, challengerName := "openshift", "basic-challenge"
 				// For now, all password challenges share a single basic challenger, since they'll all respond to any basic credentials
-				challengers["basic-challenge"] = passwordchallenger.NewBasicAuthChallenger("openshift")
+				// The separate bootstrap user realm and challengerName allows 'oc' to distinguish between bootstrap IDP and others
+				if config.IsBootstrap(identityProvider) {
+					realm, challengerName = "openshift:kubeadmin", "kubeadmin-challenge"
+				}
+				challengers[challengerName] = passwordchallenger.NewBasicAuthChallenger(realm)
 			}
 		} else if config.IsOAuthIdentityProvider(identityProvider) {
 			oauthProvider, err := c.getOAuthProvider(identityProvider)
--- a/pkg/config/helpers.go
+++ b/pkg/config/helpers.go
@@ -37,3 +37,12 @@ func IsPasswordAuthenticator(provider osinv1.IdentityProvider) bool {
 
 	return false
 }
+
+func IsBootstrap(provider osinv1.IdentityProvider) bool {
+	switch provider.Provider.Object.(type) {
+	case *BootstrapIdentityProvider:
+		return true
+	}
+
+	return false
+}


pkg/oauthserver/auth.go
```diff
@@ -292,7 +292,7 @@ func (c *OAuthServerConfig) getAuthenticationHandler(mux oauthserver.Mux, errorH
 	multiplePasswordProviders := false
 	passwordProviderCount := 0
 	for _, identityProvider := range c.ExtraOAuthConfig.Options.IdentityProviders {
-		if config.IsPasswordAuthenticator(identityProvider) && identityProvider.UseAsLogin {
+		if (config.IsPasswordAuthenticator(identityProvider) || config.IsBootstrap(identityProvider)) && identityProvider.UseAsLogin {
```
stlaz, line 295: no need to change this line

pkg/oauthserver/auth.go
```diff
@@ -308,7 +308,7 @@ func (c *OAuthServerConfig) getAuthenticationHandler(mux oauthserver.Mux, errorH
 		}
 
 		// TODO: refactor handler building per type
-		if config.IsPasswordAuthenticator(identityProvider) {
+		if config.IsPasswordAuthenticator(identityProvider) || config.IsBootstrap(identityProvider) {
```
stlaz, line 311: no need to change this line

pkg/oauthserver/auth.go
```diff
@@ -356,7 +356,12 @@ func (c *OAuthServerConfig) getAuthenticationHandler(mux oauthserver.Mux, errorH
 			}
 			if identityProvider.UseAsChallenger {
 				// For now, all password challenges share a single basic challenger, since they'll all respond to any basic credentials
-				challengers["basic-challenge"] = passwordchallenger.NewBasicAuthChallenger("openshift")
+				// The separate "bootstrap" realm allows 'oc' to distinguish between bootstrap IDP and others
+				if config.IsBootstrap(identityProvider) {
+					challengers["basic-challenge"] = passwordchallenger.NewBasicAuthChallenger("bootstrap")
+				} else {
+					challengers["basic-challenge"] = passwordchallenger.NewBasicAuthChallenger("openshift")
+				}
```
stlaz, line 364: I'd prefer this:
```golang
realm := "openshift"
if config.IsBootstrap(identityProvider) {
    realm = "openshift:kubeadmin"
}
challengers["basic-challenge"] = passwordchallenger.NewBasicAuthChallenger(realm)
```

stlaz (changes requested):

pkg/oauthserver/auth.go
sallyom: fixed, originally i took the bootstrap out of the IsPassword cases :facepalm: thanks

pkg/oauthserver/auth.go
sallyom: fixed, thanks

pkg/oauthserver/auth.go
sallyom: thanks, updated

stlaz: actually, it should probably be
```golang
realm, challengerName := "openshift", "basic-challenge"
if config.IsBootstrap(identityProvider) {
    realm, challengerName = "openshift:kubeadmin", "kubeadmin-challenge"
}
challengers[challengerName] = passwordchallenger.NewBasicAuthChallenger(realm)
```
my bad

sallyom: updated

sallyom: curious, do we use/process that "basic-challenge"/"kubeadmin-challenge" anywhere? I don't see it.

stlaz: the challengers are passed into a union here: https://github.com/openshift/oauth-server/blob/a7f975a5f3dc07f270dc64275033a66308d578a2/pkg/oauthserver/auth.go#L430

What I want is to send different WWW-Authenticate headers for kubeadmin and for normal OpenShift login, but in that case we cannot overwrite the `basic-challenge` challenger with a different one

sallyom: i see now, thanks

stlaz:
/hold
this probably was not such a good idea after all - https://github.com/openshift/oc/pull/553#discussion_r489212922

sallyom:
going in a different direction, making the change only in oc, closing this

PR #59, opened 2020-11-03
bumps to kube 1.19.2 and picks up our carries to apiserver package

p0lyn0mial:
bringing `openshift-apiserver-4.7-kubernetes-1.19.2` is important because it sets the default timeout for the SAR client


stlaz:
/lgtm
not a big fan of the kube-apiserver replace in this repo, I wonder if there's any chance upstream might accept the patch instead

p0lyn0mial:
> not a big fan of the kube-apiserver replace in this repo, I wonder if there's any chance upstream might accept the patch instead

yes, we could try to drop it on the next release.

PR #64, opened 2020-12-07
Bug 1905195: Detecting broken connections to the Kube API takes up to 15 minutes

p0lyn0mial:
we will carry a patch that allows the client to detect broken connections to the api server more quickly than the default TCP timeout of 15 minutes.

this patch will be dropped in future releases once http2/ping is introduced.

ref: https://github.com/openshift/library-go/pull/945


go.mod
```diff
@@ -32,4 +32,5 @@ require (
 replace (
 	github.com/RangelReale/osin => github.com/openshift/osin v1.0.1-0.20180202150137-2dc1b4316769
 	github.com/RangelReale/osincli => github.com/openshift/osincli v0.0.0-20160924135400-fababb0555f2
+        k8s.io/apiserver => github.com/openshift/kubernetes-apiserver  v0.0.0-20201207110950-476028df0fd8 // points to openshift-apiserver-4.6-kubernetes-1.19.0
```
sttts: formatted correctly?

p0lyn0mial: yes, otherwise go mod tidy would complain.

PR #71, opened 2021-03-25
Bug 1958080: CONSOLE-2535: Internationalize login page

jhadvig:
@sg00dwin the locales will be available in `locales.go` where all the locales will be living in a form of key-value struct. They are hardcoded into the oauth-server since there is just a handful of them. We need to update all the login related templates so they contain the go template macros.

Had to add `golang.org/x/text/language` package to get helpers for parsing the [Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) header.

/assign @stlaz


sg00dwin:
@jhadvig  
Related pr https://github.com/openshift/oauth-templates/pull/15

jhadvig:
@stlaz we are ready for review. PTAL

pkg/server/login/locales.go
```diff
@@ -0,0 +1,56 @@
+package login
+
+import "golang.org/x/text/language"
+
+func getLocale(langBase string) map[string]string {
+	switch langBase {
+	case language.English.String():
+		return locale_en
+	// Uncomment once given locale is available
```
stlaz, line 9: When is that going to be?

pkg/server/login/locales.go
```diff
@@ -0,0 +1,56 @@
+package login
+
+import "golang.org/x/text/language"
+
+func getLocale(langBase string) map[string]string {
+	switch langBase {
+	case language.English.String():
```
stlaz, line 7: How do I know this works?

pkg/server/login/login.go
```diff
@@ -128,6 +130,8 @@ func (l *Login) handleLoginForm(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	form.Locale = getLocale(getPrefferedLang(req.Header.Get("Accept-Language")))
```
stlaz, line 133: What if I want to specify the language myself, manually? That's actually very often the case for myself, I default my system to English language, but prefer websites on the Czech web to be in Czech, which oftentimes I have to do manually on Czech pages since the autodetection is wrong.

I would expect some kind of this + a language button choice of which is stored in a cookie.

stlaz (changes requested):
Our repositories require enhancements in order to merge new features. Can you please link an enhancement describing this change? It would make it simpler to reason about the alternatives to just getting "Accept-Language" header and about the timing of the translations.

pkg/server/login/locales.go
jhadvig: Tested these methods with various  Accept-Language header locally and it was working. Not sure why?

pkg/server/login/locales.go
jhadvig: Thats up to translations team that is providing those.
@sg00dwin any estimations ?

pkg/server/login/login.go
jhadvig: > What if I want to specify the language myself, manually

if we have locale for that lang it will be used, if not we default to English

>I would expect some kind of this + a language button choice of which is stored in a cookie.

This story only contains reading the lang preferences from the header and parsing the go template with appropriate locale.

jhadvig:
@stlaz thanks for review. 
@spadgett shall we incorporate an enhancement into the story ?

pkg/server/login/locales.go
mtrmac: Full-time translators typically use specialized file formats/tools/editors, a repository of previously-translated strings, and the like. How is this going to interoperate with such workflows?

Is this really such a unique and special case that asking translators to provide Go patches, or setting up a special process to manually turn contributed translations into Go patches, is worthwhile and practical?

pkg/server/login/login.go
spadgett: @stlaz We have this for console itself. The problem is that it's stored as a user preference, which is tied to a specific user. Here you haven't logged in yet, so we can't load your user preferences.

We could also store the last selected language in localStorage and pass it as a query parameter to the login page, however. Then the query parameter would override `Accept-Language` if present. @jhadvig Do you might also checking a `lng` query parameter (to match what we support in console) first, then falling back to `Accept-Language`?

Note that "preferred" is misspelled here :)

spadgett:
@stlaz We can update our existing i18n enhancement to cover the login page:

https://github.com/openshift/enhancements/blob/master/enhancements/console/internationalization.md

spadgett:
@jhadvig What are the messages we need? We can reach out to the globalization team to try to get them right away. We might already have "Username" and "Password" from other parts of the UI.

pkg/server/login/locales.go
spadgett: @mtrmac We have these tools in place for the main console repo. We thought that would be overkill here for a handful of messages that rarely change. We were planning to include whatever strings we need with the other console strings, and the console team would manage patching them when needed.

jhadvig:
@spadgett its just a handful of strings, check https://github.com/openshift/oauth-server/blob/6c96a146c6c5cab0293c37a9aa9e147b2250328a/pkg/server/login/locales.go#L41-L50

> We might already have "Username" and "Password" from other parts of the UI.

I would rather have all of the string translated before enabling locale.

spadgett:
@jhadvig I think we'll need to use i18next-parser comments in the console code base so that these messages are added automatically for translation.

```ts
// t('Log in to your account')
// t('Username')
// ... etc ...
```

Then our existing process for sending messages for translation will work. It would be a manual update to this repo when something changes, but I think that would be rare and something the console team can take on if @stlaz and @mtrmac agree. We should add this detail to our existing i18n enhancement.

We have "Username", "Password", and "Error" translated already, but not the other strings.

mtrmac:
FWIW I’m just a curious observer. I just wanted to make sure that a viable process of contributing translations actually exists before settling on this mechanism.

spadgett:
cc @rebeccaalpert

pkg/server/login/templates.go
```diff
@@ -264,26 +269,26 @@ select.pf-c-form-control.pf-m-success { --pf-c-form-control--PaddingRight: var(-
...
               <div class="pf-c-form__group pf-m-action">
-                <button class="pf-c-button pf-m-primary pf-m-block" type="submit" tabindex="3">Log in</button>
+                <button class="pf-c-button pf-m-primary pf-m-block" type="submit" tabindex="3">{{ .Locale.LogIn }}</button>
               </div>
             </form>
           </div>
         </main>
         <footer class="pf-c-login__footer">
-          <p>Welcome to OKD.</p>
+          <p>{{ .Locale.WelcomeTo }} OKD</p>
```
spadgett, line 291: I think we lost the product name here.

jhadvig: did we ? I still see the `OKD` there, we just pulled it out of the double curly brackets.

spadgett: Sorry, I missed that. Really you're not supposed to concat strings, instead making them a parameter in the message so the order of the words can change. That might be tricky here.

pkg/server/login/locales.go
sg00dwin: > Thats up to translations team that is providing those.
> @sg00dwin any estimations ?

The translation team doesn't provide estimated turn around times. Still waiting.

rebeccaalpert:
@spadgett - @sgoodwin had sent these strings over to me and I'd manually created a PO file and sent it over to Terry's team (you were CCed on that one). They assigned it out and said they'd give it back to us with the other strings from the last sprint.

If we can get the parser to pick these up, that would be great!

pkg/server/selectprovider/templates.go
```diff
@@ -188,23 +193,21 @@ body { background-color: var(--pf-global--BackgroundColor--dark-100); }
...
-              <a class="pf-c-button pf-m-primary pf-m-block" href="{{ (index .Providers 0).URL }}">Log In</a>
+              <a class="pf-c-button pf-m-primary pf-m-block" href="{{ (index .Providers 0).URL }}">{{ .Locale.LogIn }}</a>
             {{ else }}
-              <h1 class="pf-c-title pf-m-3xl">Log in with&hellip;</h1>
+              <h1 class="pf-c-title pf-m-3xl">{{ .Locale.LogInWith }}</h1>
               <ul>
                 {{ range $provider := .Providers }}
                   <li class="idp">
-                    <a href="{{$provider.URL}}" class="pf-c-button pf-m-secondary pf-m-block" title="Log in with {{$provider.Name}}">{{$provider.Name}}</a>
+                    <a href="{{$provider.URL}}" class="pf-c-button pf-m-secondary pf-m-block" title="{{ .Locale.LogInWith }} {{$provider.Name}}">{{$provider.Name}}</a>
```
rebeccaalpert, line 202: The translators generally need full sentences. I think this may cause issues because it's broken up.

pkg/server/login/templates.go
spadgett: @rebeccaalpert That doesn't work here since it's a Go template and not using i18next :/

rebeccaalpert: Bleh. Aiko did raise this as an issue re: ordering.

sgoodwin:
HI I'm sgoodwin haha. Good work internationalizing things though!

pkg/server/selectprovider/templates.go
jhadvig: @rebeccaalpert so we will need to have atranslations for the `Log In With` string  together with all the product variations ?

rebeccaalpert: I think that's safer, yeah. Like Sam mentioned above sometimes the ordering of the sentence can change depending on the language.

rebeccaalpert:
Forgot the all important 00s. Sorry other Sam! :)

pkg/server/selectprovider/templates.go
sg00dwin: The platform name is now included with the key string for the translators 
`"Welcome to {{platformTitle}}": "Welcome to {{platformTitle}}",`
https://github.com/openshift/console/pull/8587#discussion_r609747067

rebeccaalpert: Thank you!

spadgett: @sg00dwin @rebeccaalpert I'm not sure we can make that work in this context. The OAuth server won't know which platform name to use. Branding is done by replacing the entire login page template.

We might have to use string concatenation for now since we don't have the i18next library here.

pkg/server/login/templates.go
sg00dwin: > Sorry, I missed that. Really you're not supposed to concat strings, instead making them a parameter in the message so the order of the words can change. That might be tricky here.

@jhadvig how will we this need to be structured to handle translation changes to word order? eg. if  locale_jp becomes `OKD {{ .Locale.WelcomeTo }}`

pkg/server/login/locales.go
```diff
@@ -0,0 +1,63 @@
...
+	tag, _, _ := matcher.Match(userPrefs...)
+	base, _ := tag.Base()
+	return base.String()
+}
+
+var supportedLangs = []language.Tag{
+	language.English, // en - first language is fallback
+	// Uncomment once given locale is available
+	// language.Chinese,  // zh
+	// language.Japanese, // jp
```
spadgett, line 37: `ja` instead of `jp`?

pkg/server/login/locales.go
```diff
@@ -0,0 +1,63 @@
+package login
+
+import "golang.org/x/text/language"
+
+func getLocale(langBase string) map[string]string {
+	switch langBase {
+	case language.English.String():
```
spadgett, line 7: Does this work with values like `en-US` and `ja-JP` where a region is used?

jhadvig: Yes, that's done by `getPreferredLang()`

jhadvig:
@spadgett comments addressed

pkg/server/login/locales.go
sg00dwin: Translated strings are now included for each language.

jhadvig:
@stlaz I've added additional unit tests and updated the current ones. PTAL :)

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
+	golang.org/x/text v0.3.4
```
stlaz, line 23: Get rid of the JIRA reference in the commit message.

stlaz (commented):
please squash the non-auto-generated commits

pkg/server/errorpage/templates.go
```diff
@@ -9,7 +9,12 @@ var defaultErrorPageTemplate = template.Must(template.New("defaultErrorPageTempl
 const defaultErrorPageTemplateString = `<!DOCTYPE html>
 <html lang="en-us" data-test-id="login">
   <head>
-    <title>Error - OKD</title>
+    <title>
+
+        {{ .Locale.Error }}
+
+       · OKD
+    </title>
```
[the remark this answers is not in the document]
jhadvig: @sgoodwin ^^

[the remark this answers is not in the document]
sg00dwin: Whitespace, which was the browser generated source from logic here https://github.com/openshift/oauth-templates/pull/15/files#diff-e241bda4e3c3c6dc1c0b00185b61f6ce19b5eb16e294dd955ca9fa6d01befb0eR2-R9 

dot should be a dash that was unintentionally changed.

@jhadvig pr to address these. https://github.com/jhadvig/oauth-server/pull/5

jhadvig:
@stlaz comments addressed.

pkg/server/locales/locales.go
```diff
@@ -0,0 +1,78 @@
...
+	default:
+		return locale_en
+	}
+}
+
+func getPreferredLang(acceptLangHeader string) string {
+	matcher := language.NewMatcher(supportedLangs)
+	userPrefs, _, err := language.ParseAcceptLanguage(acceptLangHeader)
+	if err != nil {
+		// if error occurs, fallback to English
```
stlaz, line 25: dump this error to debug logs

pkg/server/locales/locales.go
```diff
@@ -0,0 +1,78 @@
...
+		return language.English.String()
+	}
+	tag, _, _ := matcher.Match(userPrefs...)
+	base, _ := tag.Base()
```
stlaz, line 29: no language variations?

pkg/server/locales/locales.go
```diff
@@ -0,0 +1,78 @@
+package locales
+
+import "golang.org/x/text/language"
+
+func GetLocale(acceptLangHeader string) map[string]string {
+	langBase := getPreferredLang(acceptLangHeader)
```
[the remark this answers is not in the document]
jhadvig: we can 👍

jhadvig:
@stlaz thanks for review, comments addressed 👍

pkg/server/locales/locales.go
```diff
@@ -0,0 +1,82 @@
...
+		return locale_en
+	}
+	return locale
+}
+
+func getPreferredLang(acceptLangHeader string) string {
+	matcher := language.NewMatcher(supportedLangs)
+	userPrefs, _, err := language.ParseAcceptLanguage(acceptLangHeader)
+	if err != nil {
+		klog.Warningf("Error parsing 'Accept-Language' header, falling back to English language: %v", err)
```
stlaz, line 29: debug log please (lower verbosity)

PR #73, opened 2021-03-31
Remove the legacy Dockerfile

stlaz:
The `images/Dockerfile` is unused and the builder image refers to a decommissioned server.


jhadvig (approved):
/lgtm

PR #74, opened 2021-04-08
oidc handler: add remote OIDC URLs to error logs

stlaz:
/assign @deads2k


pkg/oauth/external/handler.go
```diff
@@ -122,7 +122,7 @@ func (h *Handler) AuthenticatePassword(ctx context.Context, username, password s
 			// An invalid_grant error means the username/password was rejected
 			return nil, false, nil
 		}
-		klog.V(4).Infof("Error getting access token using resource owner password grant: %v", err)
+		klog.V(4).Infof("Error getting access token from an external OIDC provider (%s) using resource owner password grant: %v", accessReq.GetTokenUrl(), err)
```
deads2k, line 125: Can we bump to v(2)?  these should be rare unless something is misbehaving, right?

stlaz: If 100 people desperately try to login during an OIDC provider outage, this could happen more than often, but I think it should still be more discoverable than on V(4)

PR #83, opened 2021-09-14
Updating oauth-server images to be consistent with ART

s-urbaniak:


s-urbaniak:
@hardys PTAL, i think this the ci-operator yaml shold be good by now.

PR #87, opened 2021-09-24
WIP: AUTH-8: Add group synchronization from OIDC providers

stlaz:
This PR adds group synchronization for OIDC providers as described in the [Synchronizing User Groups on Login](https://github.com/openshift/enhancements/blob/master/enhancements/authentication/synchronizing-user-groups-on-login.md) enhancement.

The usermappers normally used to create users are wrapped by a delegator that first attempts to perform the usermapping strategy and only after its success it attempts to add/remove the user in appropriate groups.


stlaz:
/hold
for some reason the bot did not mark the PR as WIP, adding hold manually

stlaz:
/hold cancel
there we go

pkg/groupmapper/groupmapper.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	var newUsers []string
+	switch userIdx {
+	case -1:
+		return nil
+	case 0:
+		newUsers = updatedGroup.Users[1:]
+	default:
+		newUsers = append(updatedGroup.Users[0:userIdx], updatedGroup.Users[userIdx+1:]...)
```
ibihim, line 155: Can it happen, that the user is the last one? Would it be out of bounds?

pkg/groupmapper/groupmapper.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	return nil
+}
+
+// FIXME: work better with annotations
+// FIXME: add retries
+// FIXME: use idpName
+func (m *UserGroupsMapper) removeUserFromGroupWithRetries(idpName, username, group string) error {
+	updatedGroup, err := m.groupsClient.Get(context.TODO(), group, metav1.GetOptions{})
```
ibihim, line 123: Shouldn't we bubble up the ctx?

pkg/groupmapper/groupmapper.go
```diff
@@ -0,0 +1,207 @@
...
+			return nil
+		}
+		return err
+	}
+
+	if len(updatedGroup.Users) == 0 {
+		return nil
+	}
+
+	if len(updatedGroup.Users) == 1 && updatedGroup.Users[0] == username && updatedGroup.Annotations[groupGeneratedKey] == "true" {
```
ibihim, line 135: Nit: This line is quite long. Should we consider adding a linter to the repository?

pkg/groupmapper/groupmapper.go
```diff
@@ -0,0 +1,207 @@
...
+	return err
+}
+
+func groupsDiff(existing []*userv1.Group, required sets.String) (toRemove, toAdd []string) {
+	existingNames := sets.NewString()
+	for _, g := range existing {
+		existingNames.Insert(g.Name)
+	}
+
+	return existingNames.Difference(required).UnsortedList(), required.Difference(existingNames).UnsortedList()
```
ibihim, line 206: nice

pkg/groupmapper/groupmapper.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	updatedGroup.Users = newUsers
+
+	_, err = m.groupsClient.Update(context.TODO(), updatedGroup, metav1.UpdateOptions{})
+	return err
+}
+
+// FIXME: work better with annotations
+// FIXME: add retries
```
ibihim, line 165: They methods look quite similar, especially if we add more common logic like retries:
```
withRetries:
- Get data
- Modify data
- Update data
```

WDYT?

ibihim: Btw, couldn't we hand the `updatedGroup` down and save two requests in case there are remove AND addings?

ibihim (commented):
Great PR. I can't yet understand everything, but I hope my comments help you nevertheless.

pkg/cmd/oauth-server/server.go
```diff
@@ -33,7 +33,7 @@ func RunOsinServer(osinConfig *osinv1.OsinServerConfig, stopCh <-chan struct{})
 		return errors.New("osin server requires non-empty oauthConfig")
 	}
 
-	oauthServerConfig, err := newOAuthServerConfig(osinConfig)
+	oauthServerConfig, postStartHooks, err := newOAuthServerConfig(osinConfig)
```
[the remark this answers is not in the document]
stlaz: This is the same how the `postStartHooks` are being passed through the other apiservers (oauth-, openshift- and kube-apiservers).

I don't think return of `postStartHooks` is actually that surprising here. You're generating configuration for the apiserver, and you want the apiserver to run certain actions that make the configuration work properly.

pkg/api/types.go
```diff
@@ -42,6 +46,10 @@ type UserIdentityMapper interface {
 	UserFor(identityInfo UserIdentityInfo) (user.Info, error)
 }
 
+type GroupIdentityMapper interface {
+	GroupsFor(identityInfo UserIdentityInfo) (user.Info, error)
```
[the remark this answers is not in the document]
stlaz: this is remains of my previous implementation, I thought I deleted all. This should now be removed.

pkg/groupmapper/groupmapper.go
stlaz: golang appears to be returning an empty slice on `[tooGreatOfAnIndex:]`

pkg/groupmapper/groupmapper.go
stlaz: the context is unfortunately not wired through the oauth-server, and I fear that to really do it properly we would have to also implement https://github.com/openshift/osin/issues/82

pkg/groupmapper/groupmapper.go
stlaz: I wouldn't. I'd say it's still somewhat ok. Generally there are no characters per line imposed in our code although I usually try to use shorter lines where possible and where it makes sense.

pkg/groupmapper/groupmapper.go
stlaz: Alternative approach - use a cache of all groups and use it to do the first `Get`.

It will allow us to keep a unified approach to both the removal and addition actions and we'll avoid the live `Get`.

stlaz: done in the `use lister to get rid off unnecessary GETs` commit

I agree that the removal/addition methods are quite similar although abstracting may not prove that useful because:
1. errors during gets are handled differently
2. the API actions are way different

So at best we would have a "retry" wrapper that needs to be supplied with this custom logic compared to just adding retries to the specific API requests that are a part of this logic.

pkg/cmd/oauth-server/server.go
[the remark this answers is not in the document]
ibihim: If it is idiomatic, let keep doing it that way :smile:

stlaz:
/test verify
looks suspiciously like failed infra

pkg/oauthserver/auth.go
```diff
@@ -667,6 +697,33 @@ func (c *OAuthServerConfig) getAuthenticationRequestHandler() (authenticator.Req
...
+func newIdentityUserMapperWithGroups(
+	identities userclient.IdentityInterface,
+	users userclient.UserInterface,
+	groupsInformer userinformer.GroupInformer,
+	groups userclient.GroupInterface,
+	groupsLister userlisterv1.GroupLister,
+	userIdentityMapping userclient.UserIdentityMappingInterface,
+	method identitymapper.MappingMethodType,
+) (api.UserIdentityMapper, error) {
+	userMapper, err := identitymapper.NewIdentityUserMapper(
```
s-urbaniak, line 709: as far as i see `NewIdentityUserMapper` is unused, is there any reason to keep this separate constructor? I think we are instanciating with groups only, no?

pkg/oauthserver/oauth_apiserver.go
```diff
@@ -43,7 +49,7 @@ func init() {
 
 // TODO we need to switch the oauth server to an external type, but that can be done after we get our externally facing flag values fixed
 // TODO remaining bits involve the session file, LDAP util code, validation, ...
-func NewOAuthServerConfig(oauthConfig osinv1.OAuthConfig, userClientConfig *rest.Config, genericConfig *genericapiserver.RecommendedConfig) (*OAuthServerConfig, error) {
+func NewOAuthServerConfig(oauthConfig osinv1.OAuthConfig, userClientConfig *rest.Config, genericConfig *genericapiserver.RecommendedConfig) (*OAuthServerConfig, map[string]genericapiserver.PostStartHookFunc, error) {
```
ibihim, line 52: It would be cool if we could at some point trim down the function :smile:

pkg/groupmapper/groupmapper.go
stlaz: I'd rather stick to the smallest number of cases possible and not write code that does not need to exist

pkg/groupmapper/groupmapper.go
```diff
@@ -0,0 +1,220 @@
...
+func (m *UserGroupsMapper) addUserToGroup(idpName, username, group string) error {
+	updatedGroup, err := m.groupsLister.Get(group)
+	if errors.IsNotFound(err) {
+		_, err = m.groupsClient.Create(context.TODO(),
+			&userv1.Group{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: group,
+					Annotations: map[string]string{
+						fmt.Sprintf(groupSyncedKeyFmt, idpName): "synced",
+						groupGeneratedKey:                       "true",
```
ibihim, line 179: Nit:

Sorry, I just now realized this. Would it make sense to make `"synced"` and `"true"` constants?

PR #96, opened 2022-01-11
Bug 2038691: groupsync: create annotations if not present before writing to them

stlaz:
/assign @s-urbaniak 
/cc @openshift/openshift-team-auth-maintainers


ibihim (approved):
/lgtm

But it would be better if `Get` would return a valid default value on non-error.

PR #102, opened 2022-03-11
Update the osin dependency

arjunrn:
The updated vendored dependency contains the following changes:

1. [Reworked redirect uri validation, resolving uri references to base url, and returning the resolved url to the requesting
   function](https://github.com/openshift/osin/commit/06cbf43fd9bda76ed124a73e1c44416f7550b226).
2. [URL decode the client_id and client_secret in Authz header](https://github.com/openshift/osin/pull/175).
3. [Changing GET params to FormValue](https://github.com/openshift/osin/pull/180).
4. [Ignore port when the redirect URI is a loopback address](https://github.com/openshift/osin/pull/210).

The signature of the `ValidateUris` function was changed in [1], so changed it's usage as well.

Also, changed dependency from `RangelReale/osincli` to `openshift/osincli`. This repository uses the the forked dependency openshift/osincli with a replace directive. But the dependency openshift/osin uses the forked dependency directly. This causes the `verify-deps` target to fail. So all references to `RangelReale/osincli` have been changed to `openshift/osincli`.


pkg/server/logout/logout.go
```diff
@@ -62,5 +62,6 @@ func (l *logout) isValidRedirect(then string) bool {
 		return true
 	}
 
-	return osin.ValidateUri(l.redirect, then) == nil
```
stlaz, line 65: move the non-generated code to a separate commit, please

go.mod
```diff
@@ -112,7 +112,6 @@ require (
 )
 
 replace (
-	github.com/RangelReale/osin => github.com/openshift/osin v1.0.1-0.20180202150137-2dc1b4316769
```
pierreprinetti, line 115: If you want, in your latest commit (`Updated osin to fix the uri validate issue `), you could fix the import of `osin` as well in an explicit form. Here's how I would go about it:

```shell
find pkg cmd -name '*.go' -exec sed -i 's|"github.com/RangelReale/osin"|"github.com/openshift/osin"|' {} +
go get github.com/openshift/osin@0f4d38c6e53f
go mod tidy && go mod vendor
```

go.mod
```diff
@@ -112,7 +112,6 @@ require (
...
-	github.com/RangelReale/osincli => github.com/openshift/osincli v0.0.0-20160924135400-fababb0555f2
+	github.com/RangelReale/osin => github.com/openshift/osin v1.0.2-0.20220317075346-0f4d38c6e53f
```
pierreprinetti, line 115: If you want, in your latest commit (`Updated osin to fix the uri validate issue `), you could fix the import of `osin` as well in an explicit form. Here's how I would go about it:

```shell
find pkg cmd -name '*.go' -exec sed -i 's|"github.com/RangelReale/osin"|"github.com/openshift/osin"|' {} +
go get github.com/openshift/osin@0f4d38c6e53f
go mod tidy && go mod vendor
```

s-urbaniak: I think this definitely makes sense as the upstream repository is not maintained anymore and to the best of my knowledge our downstream fork is more or less the only one being maintained, cc @stlaz for verification.

s-urbaniak: Generally though this change is independent from the contribution being done here.

pierreprinetti: >this change is independent from the contribution being done here

Reopened #104. PTAL!

slaskawi (approved):
/lgtm
/approve

go.mod
pierreprinetti: (this comment is resolved, apologies for the noise!)

PR #103, opened 2022-03-18
Bug 2086465: External OAuth: create audit logs for auth events

pierreprinetti:
With this patch, authentication mediated by external identity providers
is logged in the audit trail. When available, the internal mapping of
the identity confirmed by the external identity provider is attached to
both successful and unsuccessful authentication events.

Implements [AUTH-70](https://issues.redhat.com/browse/AUTH-70)

### DIFF
--- a/pkg/oauth/external/github/github_test.go
+++ b/pkg/oauth/external/github/github_test.go
@@ -1,0 +1,205 @@
+package github
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/RangelReale/osincli"
+	"github.com/openshift/oauth-server/pkg/api"
+)
+
+func TestGetUserIdentity(t *testing.T) {
+	newGithubIdentityProvider := func(username string, orgs []string) http.RoundTripper {
+		return roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+			body := new(bytes.Buffer)
+
+			switch req.URL.Path {
+			case "/user":
+				if err := json.NewEncoder(body).Encode(struct {
+					ID                 uint64
+					Login, Email, Name string
+				}{
+					ID:    12345,
+					Login: username,
+					Email: "user@example.com",
+				}); err != nil {
+					panic(err)
+				}
+				return &http.Response{
+					StatusCode: http.StatusOK,
+					Status:     http.StatusText(http.StatusOK),
+					Body:       io.NopCloser(body),
+				}, nil
+
+			case "/user/orgs":
+				type ghOrg struct {
+					ID    uint64
+					Login string
+				}
+				ghOrgs := make([]ghOrg, len(orgs))
+				for i := range orgs {
+					ghOrgs[i] = ghOrg{
+						ID:    999,
+						Login: orgs[i],
+					}
+
+				}
+
+				if err := json.NewEncoder(body).Encode(ghOrgs); err != nil {
+					panic(err)
+				}
+				return &http.Response{
+					StatusCode: http.StatusOK,
+					Status:     http.StatusText(http.StatusOK),
+					Body:       io.NopCloser(body),
+				}, nil
+
+			default:
+				return nil, fmt.Errorf("this fixture does not serve the requested path: %s", req.URL.Path)
+			}
+		})
+	}
+
+	type checkFunc func(api.UserIdentityInfo, error) error
+	hasUsername := func(want string) checkFunc {
+		return func(userIdentityInfo api.UserIdentityInfo, _ error) error {
+			if have := userIdentityInfo.GetProviderPreferredUserName(); want != have {
+				return fmt.Errorf("expected username %q, got %q", want, have)
+			}
+			return nil
+		}
+	}
+	hasUsernameInError := func(want string) checkFunc {
+		return func(_ api.UserIdentityInfo, err error) error {
+			var userIdentityInfo api.UserIdentityInfo
+
+			var denied api.AuthorizationDeniedError
+			if errors.As(err, &denied) {
+				userIdentityInfo = denied.Identity()
+			}
+
+			var failed api.AuthorizationFailedError
+			if errors.As(err, &failed) {
+				userIdentityInfo = failed.Identity()
+			}
+
+			if have := userIdentityInfo.GetProviderPreferredUserName(); want != have {
+				return fmt.Errorf("expected username %q, got %q", want, have)
+			}
+			return nil
+		}
+	}
+	isAllowed := func(_ api.UserIdentityInfo, err error) error {
+		if err != nil {
+			return fmt.Errorf("unexpected error: %v", err)
+		}
+		return nil
+	}
+	isDenied := func(_ api.UserIdentityInfo, err error) error {
+		if err == nil {
+			return fmt.Errorf("expected authorization error, got nil")
+		}
+		var authError api.AuthorizationDeniedError
+		if !errors.As(err, &authError) {
+			return fmt.Errorf("expected authorization error, got %v", err)
+		}
+		return nil
+	}
+
+	for _, tc := range [...]struct {
+		name                 string
+		username             string
+		userOrganizations    []string
+		allowedOrganizations []string
+
+		checks []checkFunc
+	}{
+		{
+			name:     "ok",
+			username: "hello",
+			checks: []checkFunc{
+				isAllowed,
+				hasUsername("hello"),
+			},
+		},
+		{
+			name:                 "ok with organization check",
+			username:             "hello",
+			userOrganizations:    []string{"vip"},
+			allowedOrganizations: []string{"vip"},
+			checks: []checkFunc{
+				isAllowed,
+				hasUsername("hello"),
+			},
+		},
+		{
+			name:                 "ok with organization check, member of multiple",
+			username:             "hello",
+			userOrganizations:    []string{"openshift", "gophercloud", "vip", "kubernetes"},
+			allowedOrganizations: []string{"vip"},
+			checks: []checkFunc{
+				isAllowed,
+				hasUsername("hello"),
+			},
+		},
+		{
+			name:                 "ok with organization check, multiple allowed",
+			username:             "hello",
+			userOrganizations:    []string{"vip"},
+			allowedOrganizations: []string{"vip", "openshift", "kubernetes"},
+			checks: []checkFunc{
+				isAllowed,
+				hasUsername("hello"),
+			},
+		},
+		{
+			name:                 "denied, not in organization",
+			username:             "hello",
+			userOrganizations:    []string{"microsoft"},
+			allowedOrganizations: []string{"neovim"},
+			checks: []checkFunc{
+				isDenied,
+				hasUsernameInError("hello"),
+			},
+		},
+		{
+			name:                 "denied, not in any organization",
+			username:             "hello",
+			allowedOrganizations: []string{"vip"},
+			checks: []checkFunc{
+				isDenied,
+				hasUsernameInError("hello"),
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			userIdentityInfo, err := NewProvider(
+				"git-tub",
+				"my_client_id",
+				"my_client_secret",
+				"",
+				newGithubIdentityProvider(tc.username, tc.userOrganizations),
+				tc.allowedOrganizations,
+				nil,
+			).GetUserIdentity(&osincli.AccessData{})
+
+			for _, check := range tc.checks {
+				if e := check(userIdentityInfo, err); e != nil {
+					t.Error(e)
+				}
+			}
+		})
+	}
+
+}
+
+type roundTripperFunc func(*http.Request) (*http.Response, error)
+
+func (rt roundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return rt(req)
+}
--- a/pkg/oauth/external/handler.go
+++ b/pkg/oauth/external/handler.go
@@ -14,7 +14,9 @@ import (
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/user"
 
+	"github.com/openshift/oauth-server/pkg/api"
 	authapi "github.com/openshift/oauth-server/pkg/api"
+	"github.com/openshift/oauth-server/pkg/audit"
 	openshiftauthenticator "github.com/openshift/oauth-server/pkg/authenticator"
 	"github.com/openshift/oauth-server/pkg/authenticator/identitymapper"
 	"github.com/openshift/oauth-server/pkg/oauth/handlers"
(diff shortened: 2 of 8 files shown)


pkg/oauth/external/github/github_test.go
```diff
@@ -0,0 +1,186 @@
...
+	isDenied := func(_ api.UserIdentityInfo, have bool, _ error) error {
+		if have {
+			return fmt.Errorf("expected denied, got allowed")
+		}
+		return nil
+	}
+	hasUserID := func(want string) checkFunc {
+		return func(userIdentityInfo api.UserIdentityInfo, _ bool, _ error) error {
+			if have := userIdentityInfo.GetIdentityName(); want != have {
+				return fmt.Errorf("expected username %q, got %q", want, have)
```
ibihim, line 86: the `%q` thing is pretty nice

pkg/oauth/external/github/github_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/RangelReale/osincli"
+	"github.com/openshift/oauth-server/pkg/api"
+)
+
+func TestGetUserIdentity(t *testing.T) {
```
ibihim, line 15: Nit: In general I like your code very much, but I don't like to put everything into the TestFunc. I would hope that a lot of the code can be reused later on :smile:

pkg/oauth/external/handler_test.go
```diff
@@ -149,3 +247,71 @@ func TestRedirectingStateError(t *testing.T) {
...
+func (attributes) GetVerb() string         { return "" }
+func (attributes) IsReadOnly() bool        { return true }
+func (attributes) GetNamespace() string    { return "" }
+func (attributes) GetResource() string     { return "" }
+func (attributes) GetSubresource() string  { return "" }
+func (attributes) GetName() string         { return "" }
+func (attributes) GetAPIGroup() string     { return "" }
+func (attributes) GetAPIVersion() string   { return "" }
+func (attributes) IsResourceRequest() bool { return false }
+func (attributes) GetPath() string         { return "" }
```
ibihim, line 263: There is this one amazing trick that @s-urbaniak taught me, you can do the following:

```Go
type attributes struct {
  authorizer.Attributes
}
```

and then define only the methods you care about. Delegating on an interface :exploding_head:

ibihim (commented):
cool

pkg/oauth/external/github/github_test.go
```diff
@@ -0,0 +1,186 @@
...
+		return func(userIdentityInfo api.UserIdentityInfo, _ bool, _ error) error {
+			if have := userIdentityInfo.GetIdentityName(); want != have {
+				return fmt.Errorf("expected username %q, got %q", want, have)
+			}
+			return nil
+		}
+	}
+	noError := func(_ api.UserIdentityInfo, _ bool, err error) error {
+		if err != nil {
+			return fmt.Errorf("unexpected error: %v", err)
```
ibihim, line 93: `%w` can be used for `error`.

pkg/oauth/external/github/github_test.go
pierreprinetti: Nothing prevents the future developer to idiomatically copy-paste! Or worst case, to extract into a function and reuse :P

pkg/oauth/external/github/github_test.go
pierreprinetti: True, however it's useless in tests.

pkg/oauth/external/handler_test.go
pierreprinetti: Nice! Saved me a couple lines. PTAL

pkg/oauth/external/github/github_test.go
```diff
@@ -0,0 +1,201 @@
...
+				}, nil
+
+			default:
+				return nil, fmt.Errorf("this fixture does not serve the requested path: %s", req.URL.Path)
+			}
+		})
+	}
+
+	type checkFunc func(api.UserIdentityInfo, error) error
+	checks := func(fns ...checkFunc) []checkFunc { return fns }
```
stlaz, line 71: why don't you just add the functions directly to a slice?

stlaz (changes requested):

pkg/oauth/external/handler.go
```diff
@@ -180,29 +176,47 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-	if !ok {
-		klog.V(4).Infof("Could not get userIdentityInfo info from access token")
-		err := errors.New("Could not get userIdentityInfo info from access token")
-		h.handleError(err, w, req)
-		return
+		var authorizationError AuthorizationError
+		if errors.As(err, &authorizationError) {
+			klog.V(4).Infof("Authorization error: %v", authorizationError.AuthorizationDenialReason())
+
+			userInfo, err := h.mapper.UserFor(identity)
```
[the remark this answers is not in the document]
ibihim: I like the idea of having a username field in the error and not relying on the user mapping.

Right before you went on vacation, I was thinking through the general problem of getting the decision as late and the username as early as possible. It is problematic as the logic is placed on different stages.
Looking on your dedicated error solution (after you mentioned your `AuthorizationError` solution) and discussing it in our team chat, I found the solution of an error with a username field like a very good solution. A solution that could at some point create a good overall solution.

But I might be biased as I worked on this topic a little bit.

pkg/oauth/external/github/github_test.go
ibihim: it looks and reads cleaner :smile: 

I enjoy such constructs quite a lot :sweat_smile:

pkg/oauth/external/handler_test.go
```diff
@@ -149,3 +264,74 @@ func TestRedirectingStateError(t *testing.T) {
...
+
+type attributes struct {
+	authorizer.Attributes
+}
+
+func (attributes) GetUser() user.Info      { return nil }
+func (attributes) GetVerb() string         { return "" }
+func (attributes) IsResourceRequest() bool { return false }
+
+type provider struct {
```
ibihim, line 276: @s-urbaniak showed me this neat trick: https://github.com/openshift/oauth-server/pull/107/files#diff-df8eb2f6bcbdb1bd156b4feaf6a76f7879dbeae80a28ce00add3be51943a4185R284-R290

It reliefs you of the necessity to define function you won't use, just to satisfy an interface.

pkg/oauth/external/interfaces.go
```diff
@@ -27,3 +31,8 @@ type State interface {
 	Generate(w http.ResponseWriter, req *http.Request) (string, error)
 	Check(state string, req *http.Request) (bool, error)
 }
+
+type AuthorizationError interface {
```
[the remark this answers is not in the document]
pierreprinetti: done

pkg/oauth/external/handler_test.go
```diff
@@ -149,3 +264,74 @@ func TestRedirectingStateError(t *testing.T) {
...
+}
+
+func (provider) NewConfig() (*osincli.ClientConfig, error)     { return nil, nil }
+func (provider) GetTransport() (http.RoundTripper, error)      { return nil, nil }
+func (provider) AddCustomParameters(*osincli.AuthorizeRequest) {}
+func (p provider) GetUserIdentity(*osincli.AccessData) (api.UserIdentityInfo, error) {
+	return userIdentityInfo{providerName: p.name, username: p.username}, p.err
+}
+
+type userIdentityInfo struct {
```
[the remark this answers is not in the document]
pierreprinetti: done

pkg/oauth/external/handler_test.go
```diff
@@ -149,3 +264,74 @@ func TestRedirectingStateError(t *testing.T) {
...
+func (u userIdentityInfo) GetIdentityName() string {
+	return u.GetProviderName() + ":" + u.GetProviderUserName()
+}
+func (u userIdentityInfo) GetProviderName() string            { return u.providerName }
+func (u userIdentityInfo) GetProviderUserName() string        { return u.username }
+func (userIdentityInfo) GetProviderGroups() []string          { return nil }
+func (userIdentityInfo) GetExtra() map[string]string          { return make(map[string]string) }
+func (userIdentityInfo) GetProviderPreferredUserName() string { return "" }
+
+type userInfo struct {
```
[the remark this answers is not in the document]
pierreprinetti: done

pkg/oauth/external/handler_test.go
```diff
@@ -149,3 +264,74 @@ func TestRedirectingStateError(t *testing.T) {
 		t.Errorf("Expected original error back, got %#v", err)
 	}
 }
+
+type attributes struct {
+	authorizer.Attributes
```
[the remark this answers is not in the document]
pierreprinetti: done

pkg/oauth/external/handler.go
```diff
@@ -180,29 +176,47 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+				klog.V(4).Infof("Got userIdentityMapping: %#v", userInfo)
+				audit.AddUsernameAnnotation(req, userInfo.GetName())
+			} else {
+				klog.V(4).Infof("Error creating or updating mapping for: %#v due to %v", identity, err)
+			}
+
+			audit.AddDecisionAnnotation(req, audit.DenyDecision)
+			h.handleError(err, w, req)
+			return
+		} else {
```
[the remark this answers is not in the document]
pierreprinetti: done

pkg/oauth/external/handler.go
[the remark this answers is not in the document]
pierreprinetti: I'm not sure I get this one.
We decided we want to log the username _the OpenShift user was going to impersonate_, right? Isn't `mapper.UserFor` the method that returns exactly that?

Are you saying that calling `mapper.UserFor` has side-effects on the cluster?

pkg/oauth/external/github/github_test.go
pierreprinetti: done

pkg/oauth/external/handler_test.go
pierreprinetti: done

pkg/oauth/external/handler.go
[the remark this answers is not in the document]
pierreprinetti: In other terms: are you asking I re-implement the mapper logic in the `external` package so that we don't call the actual `mapper.UserFor` method?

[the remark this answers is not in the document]
stlaz: Yes, `mapper.UserFor` will actually create the identity and the user if these do not exist.

Ref: https://github.com/openshift/oauth-server/blob/9c73adb4cc72c9039cfb01b9e86b7e61000cf3be/pkg/userregistry/identitymapper/mapper.go#L33-L34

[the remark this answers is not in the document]
pierreprinetti: Added a `lookup`-method mapper to the handler, which is used for deny cases. Please take a look at the latest commit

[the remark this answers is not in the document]
pierreprinetti: Removed the additional mapper; `external.login` now uses `identity.GetProviderPreferredUserName`. PTAL

pierreprinetti:
@stlaz

As you requested, I have changed the logic so that after calling `identity, err := h.provider.GetUserIdentity(accessData)`, if `err` is non-nil, then `identity` is never accessed.

PTAL

pkg/oauth/external/errors.go
```diff
@@ -0,0 +1,33 @@
...
+	return AuthorizationDeniedError{
+		Identity: identity,
+		error:    err,
+	}
+}
+
+func (e AuthorizationDeniedError) Unwrap() error { return e.error }
+
+type AuthorizationFailedError struct {
+	Identity api.UserIdentityInfo
```
stlaz, line 22: I'd rather this was exposed with a method

pkg/oauth/external/handler_test.go
```diff
@@ -149,3 +280,39 @@ func TestRedirectingStateError(t *testing.T) {
...
+}
+
+func (p provider) GetUserIdentity(*osincli.AccessData) (api.UserIdentityInfo, error) {
+	return &api.DefaultUserIdentityInfo{
+		ProviderName:     p.name,
+		ProviderUserName: p.username,
+	}, p.err
+}
+
+type userIdentityMapper struct{}
```
stlaz, line 298: This mocking here is the reason why we did not see that requests to the API would've been made in the error case, which would've been a good red flag.

Could you mock a real mapper with fakeclients instead?

pkg/oauth/external/interfaces.go
```diff
@@ -17,8 +17,9 @@ type Provider interface {
 	GetTransport() (http.RoundTripper, error)
 	// AddCustomParameters allows an external oauth provider to provide parameters that are extension to the spec.  Some providers require this.
 	AddCustomParameters(*osincli.AuthorizeRequest)
-	// GetUserIdentity takes the external oauth token information this and returns the user identity, isAuthenticated, and error
-	GetUserIdentity(*osincli.AccessData) (authapi.UserIdentityInfo, bool, error)
+	// GetUserIdentity takes the external oauth token information and returns the user identity or a non-nil error.
+	// When error is non-nil and the user identity is available, the returned error is of type AuthorizationDenialError or AuthorizationFailureError.
```
stlaz, line 21: :+1:

pkg/oauth/external/errors.go
pierreprinetti: done

pkg/oauth/external/handler_test.go
pierreprinetti: The code now traces calls to the mapper; the test cases check that the mapper is never called upon unsuccessful auth.

pkg/oauth/external/errors.go
```diff
@@ -0,0 +1,35 @@
+package external
+
+import (
+	"github.com/openshift/oauth-server/pkg/api"
+)
+
+type AuthorizationDeniedError struct {
```
stlaz, line 7: needs godoc

pkg/oauth/external/errors.go
```diff
@@ -0,0 +1,35 @@
...
+	return AuthorizationDeniedError{
+		identity: identity,
+		error:    err,
+	}
+}
+
+func (e AuthorizationDeniedError) Identity() api.UserIdentityInfo { return e.identity }
+func (e AuthorizationDeniedError) Unwrap() error                  { return e.error }
+
+type AuthorizationFailedError struct {
```
stlaz, line 22: needs godoc

pkg/oauth/external/errors.go
```diff
@@ -0,0 +1,35 @@
+package external
```
stlaz, line 1: move to `pkg/api`

pkg/oauth/external/handler.go
```diff
@@ -180,29 +176,48 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+
 		klog.V(4).Infof("Error getting userIdentityInfo info: %v", err)
-		h.handleError(err, w, req)
-		return
-	}
-	if !ok {
-		klog.V(4).Infof("Could not get userIdentityInfo info from access token")
-		err := errors.New("Could not get userIdentityInfo info from access token")
+		audit.AddDecisionAnnotation(req, audit.ErrorDecision)
 		h.handleError(err, w, req)
```
stlaz, line 205: I think switch case would be more readable here:
```go
var authorizationDeniedError AuthorizationDeniedError
var authorizationFailedError AuthorizationFailedError
switch {
    case errors.As(err, &authorizationDeniedError):
         ....
    case errors.As(err, &authorizationFailedError):
        ...
    default:
        ...
}
```
I'll leave that up to you though

stlaz (commented):
just formalities

pkg/oauth/external/handler.go
pierreprinetti: done

pkg/oauth/external/errors.go
pierreprinetti: added

pkg/oauth/external/errors.go
pierreprinetti: moved

pkg/oauth/external/errors.go
pierreprinetti: added

PR #104, opened 2022-03-18
osin: Import from the new openshift module path

pierreprinetti:
This patch changes the import path of osin from
`github.com/RangelReale/osin` to `github.com/openshift/osin` while
keeping the dependency on the same commit.


pierreprinetti:
Duplicate of #102 
/close

pierreprinetti:
@stlaz PTAL

pierreprinetti:
This was reopened because #102 addressed the version bump, and not the import path change. This PR addresses the import path change.

PR #106, opened 2022-04-19
AUTH-100: Add request header authn auditing

ibihim:
## What

- Add audit logging to request header authentication, which is used in auth-proxy use cases.
- Add more audit logging in other areas as a catch-up.

## Why

Users want to have audit logging. A crucial flow is the request header flow.

For the catch-ups topics:
- Authentication happens in the HandleAccess case and is not audited.
- The optimistic authentication audit logging should handle all happy cases in general.

### DIFF
--- a/pkg/audit/authreq.go
+++ b/pkg/audit/authreq.go
@@ -17,11 +17,11 @@ func (f AuthenticatorFunc) AuthenticateRequest(req *http.Request) (*authenticato
 	return f(req)
 }
 
-// AuthenticatorRequestWithAuditDecision annotates the audit log with the final decision of the
+// AuthenticatorRequestWithAudit annotates the audit log with the final decision of the
 // union of authenticator.Requests. We can have several authenticator.Requesters hidden behind one
 // and unified with union. If we set the decision on anything except the last, we will get a
 // potentially wrong decision.
-func AuthenticatorRequestWithAuditDecision(authReq authenticator.Request) authenticator.Request {
+func AuthenticatorRequestWithAudit(authReq authenticator.Request) authenticator.Request {
 	return AuthenticatorFunc(func(req *http.Request) (*authenticator.Response, bool, error) {
 		res, ok, err := authReq.AuthenticateRequest(req)
 
@@ -32,6 +32,12 @@ func AuthenticatorRequestWithAuditDecision(authReq authenticator.Request) authen
 			AddDecisionAnnotation(req, DenyDecision)
 		case ok:
 			AddDecisionAnnotation(req, AllowDecision)
+		}
+
+		if res != nil && res.User != nil {
+			// This is only best effort. In the case of !ok or err, it won't have
+			// values in most cases.
+			AddUsernameAnnotation(req, res.User.GetName())
 		}
 
 		return res, ok, err
--- a/pkg/audit/authreq_test.go
+++ b/pkg/audit/authreq_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/oauth-server/pkg/audit"
 	kaudit "k8s.io/apiserver/pkg/audit"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
+	"k8s.io/apiserver/pkg/authentication/user"
 )
 
 // mustShallowRequestWithAnnotations could be extended with more features in the
@@ -32,6 +33,14 @@ func makeAuthReq(res *authenticator.Response, ok bool, err error) authenticator.
 	})
 }
 
+func makeAuthResponse(username string) *authenticator.Response {
+	return &authenticator.Response{
+		User: &user.DefaultInfo{
+			Name: username,
+		},
+	}
+}
+
 func TestAuthenticationRequestWithAuditDecision(t *testing.T) {
 	for _, tt := range [...]struct {
 		name string
@@ -40,9 +49,10 @@ func TestAuthenticationRequestWithAuditDecision(t *testing.T) {
 	}{
 		{
 			name: "should audit allow decision",
-			have: makeAuthReq(nil, true, nil),
+			have: makeAuthReq(makeAuthResponse("Franta"), true, nil),
 			want: map[string]string{
 				audit.DecisionAnnotation: string(audit.AllowDecision),
+				audit.UsernameAnnotation: "Franta",
 			},
 		},
 		{
@@ -68,7 +78,7 @@ func TestAuthenticationRequestWithAuditDecision(t *testing.T) {
 		},
 	} {
 		t.Run(tt.name, func(t *testing.T) {
-			authReq := audit.AuthenticatorRequestWithAuditDecision(tt.have)
+			authReq := audit.AuthenticatorRequestWithAudit(tt.have)
 			req := mustShallowRequestWithAnnotations(t)
 
 			_, _, _ = authReq.AuthenticateRequest(req)
--- a/pkg/authenticator/request/headerrequest/requestheader.go
+++ b/pkg/authenticator/request/headerrequest/requestheader.go
@@ -7,6 +7,7 @@ import (
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 
 	authapi "github.com/openshift/oauth-server/pkg/api"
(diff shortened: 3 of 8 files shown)


pkg/audit/authreq.go
```diff
@@ -34,6 +35,14 @@ func AuthenticatorRequestWithAuditDecision(authReq authenticator.Request) authen
 			AddDecisionAnnotation(req, AllowDecision)
 		}
 
+		if res != nil && res.User != nil {
+			// This is only best effort. In the case of !ok or err, it won't have
+			// values in most cases.
+			AddUsernameAnnotation(req, res.User.GetName())
+		} else {
+			klog.V(5).Infof("could not audit user name on authentication attempt")
```
s-urbaniak, line 43: what is the default verbosity level? we don't specify one as per https://github.com/openshift/cluster-authentication-operator/blob/master/bindata/oauth-apiserver/deploy.yaml, I believe it is 4.

what is the value of having this in logs? if it is unset in the deployment most likely this will never ever be read out. i believe introducing a metric is more sensible in this case. we don't spam logs and can correlate requests without user names.

pkg/audit/authreq_test.go
```diff
@@ -32,6 +33,21 @@ func makeAuthReq(res *authenticator.Response, ok bool, err error) authenticator.
...
+type FakeUserInfo struct {
+	username string
+	user.Info
+}
+
+func (f *FakeUserInfo) GetName() string {
+	return f.username
+}
+
+func makeAuthRes(username string) *authenticator.Response {
```
s-urbaniak, line 45: please spell this out.

pkg/oauth/handlers/authenticator.go
```diff
@@ -101,24 +102,30 @@ func (h *accessAuthenticator) HandleAccess(ar *osin.AccessRequest, w http.Respon
...
 
-		if e, ok := ar.Client.(TokenMaxAgeSeconds); ok {
-			if maxAge := e.GetTokenMaxAgeSeconds(); maxAge != nil {
-				ar.Expiration = *maxAge
-			}
+	// Disable refresh_token generation
+	ar.GenerateRefresh = false
+	ar.Authorized = true
+	if info != nil {
+		// TODO something with audiences?
```
s-urbaniak, line 122: sigh 😞

pkg/oauth/handlers/authenticator.go
```diff
@@ -101,24 +102,30 @@ func (h *accessAuthenticator) HandleAccess(ar *osin.AccessRequest, w http.Respon
 		klog.Warningf("Received unknown access token type: %s", ar.Type)
 	}
 
+	audit.AddUsernameAnnotation(ar.HttpRequest, ar.Username)
```
stlaz, line 105: aren't the annotations handled in the above authentication handling code already?

pkg/oauth/handlers/authenticator.go
ibihim: I didn't do it. I just shuffled it around. :)

pkg/authenticator/request/headerrequest/requestheader.go
```diff
@@ -49,7 +51,12 @@ func (a *Authenticator) AuthenticateRequest(req *http.Request) (*authenticator.R
 		identity.Extra[authapi.IdentityPreferredUsernameKey] = preferredUsername
 	}
 
-	return identitymapper.ResponseFor(a.mapper, identity)
+	res, ok, err := identitymapper.ResponseFor(a.mapper, identity)
+	if res != nil && res.User != nil {
+		audit.AddUsernameAnnotation(req, res.User.GetName())
```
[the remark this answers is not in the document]
ibihim: I am adding the decisions at the very last moment. You can't change a decision. So I am handling it in the middleware in auth.go

pkg/audit/authreq.go
ibihim: I'll remove it.

pkg/authenticator/request/headerrequest/requestheader.go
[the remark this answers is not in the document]
ibihim: All Authenticator.Request functions should get their decision here: https://github.com/openshift/oauth-server/blob/master/pkg/audit/authreq.go#L28-L35=
I ran into cases, where I had a deny, instead of an allow, when adding decisions down in the username logic. As mentioned, it somehow happened that the basic request decision was set to deny and you couldn't change it back to allow, once it reached requestheader.

pkg/oauth/handlers/authenticator.go
ibihim: Not really. For `authenticator.Request`, we are handling the username inside as this functions contain the logic dismantle the request and generate a username out of it. For the case of `authenticator.Password` the username and password are already known and handed down. So there is no need to have that logic within the `AuthenticatePassword` methods. Especially as there are fewer usages then implementations.

ibihim: 8 implementations (+1 in vendor, which I can't influence without changing library-go) vs 3 usages.

ibihim: And `AuthenticationAssertion` and `AuthenticateClient` don't handle username auditing afaik. But they seem to be very edge-case-y.

pkg/authenticator/request/headerrequest/requestheader.go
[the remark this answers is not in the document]
ibihim: I added a logic to `audit.AddUsernameAnnotation` to only log non-empty strings. So we can be more eager in trying to set the username.

pkg/oauth/handlers/authenticator.go
ibihim: @s-urbaniak, should we create a story for that?

s-urbaniak: @ibihim i would say so yes, at least to assert the context

pkg/authenticator/request/headerrequest/requestheader.go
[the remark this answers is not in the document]
stlaz: This looks like a design flaw. Maybe we could add the name of the provider into the decision annotation? Something like `"authentication.openshift.io/decision.<providerName>"`? Not sure what would be the simplest way to make it easily consumable by, say, `jq`.

Seeing denies/allows of all the attempted IdPs in the event seems to make a lot of sense in my head, I wonder whether you would agree.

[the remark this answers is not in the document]
ibihim: I think in combination with the OIDC IdP, where getName returns an ID, it would make sense to do so. Let me sync with @pierreprinetti on that.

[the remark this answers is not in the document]
ibihim: The design of auth-6 is kinda problematic as there are several factors that come into play:
- you can set an annotation only once,
- you have a list of IDPs, so you can't set the decision of the very first IDP,
- the logic is spread around and confusing to follow and
- on error cases we don't return the identity, just an error or false.

The best idea that arose in a chat between us, was to add a username to a custom error and get rid of the bool flag. But this could be a bigger refactoring, as we would violate the authenticator.Request interface.

pkg/authenticator/request/headerrequest/requestheader.go
```diff
@@ -34,6 +35,7 @@ func NewAuthenticator(providerName string, config *Config, mapper authapi.UserId
 func (a *Authenticator) AuthenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
 	id := headerValue(req.Header, a.config.IDHeaders)
 	if len(id) == 0 {
+		audit.AddDecisionAnnotation(req, audit.DenyDecision)
```
ibihim, line 38: @stlaz, this it then wrong, right? If the basic auth comes afterwards, it will be always deny. Let me remove it.

pkg/authenticator/request/headerrequest/requestheader.go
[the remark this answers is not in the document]
ibihim: As of our discussion, I assume there is nothing to do here.

PR #112, opened 2022-08-31
Bump library-go to b2073c41366aae78868b24e86d99021b79d754cb

alvaroaleman:
To get https://github.com/openshift/library-go/pull/1388


cmd/oauth-server/main.go
```diff
@@ -57,6 +57,7 @@ func NewOpenshiftIntegratedOAuthServerCommand(stopCh <-chan struct{}) (*cobra.Co
 			os.Exit(1)
 		},
 	}
+	logs.AddFlags(cmd.PersistentFlags())
```
alvaroaleman, line 60: I don't completely understand why this is needed, the component-base diff is this: https://github.com/kubernetes/component-base/compare/v0.22.2..v0.24.0

Component-base started doing this though, without it the global logging flags are gone, with it, global flags are the same as on `master`

stlaz: Looks like there was an upstream change.

From what I gather in https://github.com/kubernetes/component-base/commit/0c68d203462388c5319702773d799232d01def68, we should now run the `*cobra.Command` with `cli.Run()`, `cli` being `k8s.io/component-base/cli`

PR #115, opened 2023-02-06
OCPBUGS-2457: don't log request query and fragment on failed authn request

stlaz:
/assign @ibihim

### DIFF
--- a/pkg/oauth/external/handler.go
+++ b/pkg/oauth/external/handler.go
@@ -124,7 +124,7 @@ func (h *Handler) AuthenticatePassword(ctx context.Context, username, password s
 			// An invalid_grant error means the username/password was rejected
 			return nil, false, nil
 		}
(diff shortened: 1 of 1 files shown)


pkg/oauth/external/handler.go
```diff
@@ -139,6 +139,18 @@ func (h *Handler) AuthenticatePassword(ctx context.Context, username, password s
 	return identitymapper.ResponseFor(h.mapper, identity)
 }
 
+// sanitizeURL removes possible sensitive data from the URL, useful for logging
+func sanitizeURL(u *url.URL) *url.URL {
```
ibihim, line 143: Functions are copying the value, so having a `url.URL` struct as an argument instead of a pointer, would take that part off you. Except of the `URL.User`,  as you do below already.

stlaz: fixed

stlaz:
/retest-required
unrelated failures

PR #118, opened 2023-02-28
OCPBUGS-2457: don't log request query and fragment on failed authn request

stlaz:
There was one more log that was overlooken in the previous fix.

/cc @liouk


liouk (commented):
/lgtm

PR #129, opened 2023-06-22
OCPBUGS-13906: don't log request query and fragment on failed authn request 

stlaz:
Backport of https://github.com/openshift/oauth-server/pull/118/ and https://github.com/openshift/oauth-server/pull/115

+ adds images sync

/assign @ibihim


pkg/oauth/external/handler.go
```diff
@@ -139,6 +139,15 @@ func (h *Handler) AuthenticatePassword(ctx context.Context, username, password s
 	return identitymapper.ResponseFor(h.mapper, identity)
 }
 
+// sanitizeURL removes possible sensitive data from the URL, useful for logging
+func sanitizeURL(u url.URL) *url.URL {
+	u.RawQuery = ""
+	u.RawFragment = ""
+	u.User = nil
```
ibihim, line 146: Value receiver functions are only doing shallow copies. I would not expect that the root objects' `*UserInfo` gets nullified in a function called `sanitizeURL`.

ibihim (changes requested):
Great thing to add!

pkg/oauth/external/handler.go
stlaz: This is a backport, there shall be no additional changes.

The behavior matches the godoc. Also, the root object's UserInfo is not getting nullified.

ibihim: Oh sorry, I misinterpreted it.

PR #143, opened 2024-04-02
Add copy token buttons

kowalski7cc:
Enhance token web page with a copy token button.

Not really tested this because I built the image, but didn't manage to replace the used one in OpenShift.


kowalski7cc:
@ibihim @liouk is there something that can be done for this pull? It would be a nice addition 😀

PR #144, opened 2024-04-10
[4.12] OCPBUGS-27478: bump osin deps

stlaz:
/assign @liouk


xingxingxia:
Fetched the fix branch, ran grep as below for the two key functions, opened all found files and located the places of the functions, now all places indeed used safe comparing methods.
```
$ grep -lrE --exclude="*_test.go" "CheckClient...|Client...Matches"
pkg/osinserver/registrystorage/storage.go
vendor/github.com/openshift/osin/access.go
vendor/github.com/openshift/osin/client.go
vendor/github.com/openshift/osin/util.go
vendor/github.com/openshift/osin/authorize.go
```
And the e2e jobs passed. Therefore adding below qe labels:
/label qe-approved
/label cherry-pick-approved

PR #150, opened 2024-06-06
OCPBUGS-33695: p/o/handlers - fix nil-pointer, if ResStatus is nil

ibihim:
## What

Set empty ResponseStatus if nil.

## Why

- If `ResponseStatus` is nil, we have a nil pointer.
- Setting a `ResponseStatus` enables us to set the value from `getAuthMethods(req)`.

### DIFF
--- a/pkg/oauth/handlers/default_auth_handler.go
+++ b/pkg/oauth/handlers/default_auth_handler.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apiserver/pkg/audit"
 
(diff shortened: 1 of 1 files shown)


pkg/oauth/handlers/default_auth_handler.go
```diff
@@ -116,6 +117,10 @@ func (authHandler *unionAuthenticationHandler) AuthenticationNeeded(apiClient au
 				w.WriteHeader(http.StatusUnauthorized)
 				ev := audit.AuditEventFrom(req.Context())
 				if ev != nil {
+					if ev.ResponseStatus == nil {
```
stlaz, line 120: When would that happen?

Illustrate on https://github.com/openshift/oauth-server/blob/5597860a7484ea3e680b2564a4ba89a83888ace3/vendor/k8s.io/apiserver/pkg/endpoints/filters/audit.go#L42

ibihim: Hypershift is running without an audit policy file. If I drop the audit policy file in the oauth-server on OCP, I am able to reproduce the issue.

`WithAudit` isn't being turned on, without an audit policy file, but there is a function that is being run always: https://github.com/openshift/oauth-server/blob/master/vendor/k8s.io/apiserver/pkg/endpoints/filters/audit_init.go#L41-L42

ibihim:
closing in favor of #151

PR #151, opened 2024-06-07
OCPBUGS-33695: fix nil deref in default auth handler

sjenning:
https://issues.redhat.com/browse/OCPBUGS-33695


deads2k:
/approve
/hold

https://github.com/openshift/oauth-server/pull/150 looks like a better fix that retains most of the debugging info.  I've approved so @ibihim can select which fix he wants.

ibihim:
We will move forward with #150 instead.

/unhold

ibihim:
We will move forward with #150, as it keeps more information.

ibihim:
https://github.com/openshift/oauth-server/pull/150#issuecomment-2162913876

/close

PR #160, opened 2024-10-24
[WIP] OCPBUGS-43610: Disable login button upon user loggin in

jhadvig:
Adding following actions:
- disable login button when username and password are not entered
- disable login button when login button is clicked and login process takes place
- enable the login button if error occurs


logonoff:
/close

Succeeded by https://github.com/openshift/oauth-server/pull/169

PR #163, opened 2024-10-28
replace space with escape char for go 1.22

deads2k:
Need to figure out how to test this.  Maybe we just merge it in 4.18 (don't backport) and see what happens?

/assign @liouk


deads2k:
ilias opened a better otion with a test

PR #170, opened 2025-01-15
OSASINFRA-3534: Bump Gophercloud to v2

dkokkino:
Ran the script found at https://github.com/gophercloud/gophercloud/blob/main/docs/MIGRATING.md to bump gophercloud from v1 to v2

- Updated import paths to include `/v2` for Gophercloud modules
- Refactored code to account for moved and removed packages in Gophercloud v2
- Added context.TODO() where context is required but not yet implemented
- Replaced deprecated extensions with newer API structures where necessary
- Ran `go mod tidy` and re-generated vendor files to reflect updated dependencies


pierreprinetti:
@ibihim Gophercloud v1 is now frozen. The current stable branch is v2; it's functionally backwards-compatible but now accepts `context.Context` in all function signatures. PTAL!

PR #179, opened 2025-06-09
[ART-13123] append .0 to go.mod version

ashwindasr:
The go.mod change is required for ART builds. [http://github.com/openshift/build-machinery-go](https://github.com/openshift/build-machinery-go) will need to be updated to permit 1.y.z and 1.y matches before the verify job will succeed


ibihim:
You must run `make update-deps-overrides` to fix `make verify-deps`.

PR #197, opened 2025-08-22
CNTRLPLANE-1314: Update openshift/oauth-server to latest Kubernetes 1.33 version

sanchezl:
Bump kube and openshift dependencies.


pkg/server/grant/grant.go
```diff
@@ -126,7 +127,7 @@ func (l *Grant) handleForm(user user.Info, w http.ResponseWriter, req *http.Requ
 	redirectURI := q.Get(redirectURIParam)
 
 	client, err := l.clientregistry.Get(context.TODO(), clientID, metav1.GetOptions{})
-	if err != nil || client == nil {
+	if err != nil || client == nil || reflect.ValueOf(client).Elem().IsZero() {
```
sanchezl: * The new generic client-go changes the behavior of `Get` to [return a zero value instead of a nil](https://github.com/skitt/kubernetes/blob/2bdcad7419c7a584da4174513e36f867e88424df/staging/src/k8s.io/client-go/gentype/fake.go#L125-L127).
* The `OAuthClient` struct is non-comparable, so I had to resort to using the `reflect` package here.
* This change is needed for tests that use the [badClientRegistry](https://github.com/openshift/oauth-server/blob/7d49ee1475fb5a9409a2821fe34135abb8883ce0/pkg/server/grant/grant_test.go#L56-L62), which IMO is an especially bad client registry.
   * It returns `true, nil, nil` which I do not think it [is a valid result](https://github.com/openshift/library-go/blob/cf8518058622ea3fd6bb2336f24f18e8e4264303/pkg/oauth/oauthserviceaccountclient/oauthclientregistry.go#L230).
   * As I see it, it should always return a valid client when err == nil, and we should be expecting a NotFound error in the case where it is currently returning `nil, nil`.

In the end, I opted to perform the minimum needed to pass the test, more testing would be needed to fully refactor.

sanchezl: Further looking into it, this new behavior from the generic fake `Get` actually matches the [actual `Get` behavior.](https://github.com/kubernetes/kubernetes/blob/37ed906a33211c7d578cab2d681941ebfd2f2f23/staging/src/k8s.io/client-go/gentype/type.go#L160-L171).

benluddy: I agree this isn't something that is supposed to happen. Can we change the two test cases to call badClientRegistry with non-nil errors instead?  I'd prefer to see the change in the fake client behavior addressed with test-only changes.

sanchezl: I've changed the test case to match the expected behavior.

sanchezl:
/payload-with-prs 4.20 nightly informing openshift/oauth-server#197 openshift/oauth-apiserver#139

ibihim:
/retest-required

Automated retry triggered by CI triage analysis.

**Failure Analysis:**
- `ci/prow/e2e-aws-serial` failed due to 4-hour timeout during deprovision phase
- All 231 e2e tests that ran completed successfully (0 failures)
- The job exceeded time limit during infrastructure cleanup, not during test execution
- Classification: Infrastructure issue (95% confidence)

This is a transient timing issue, not a code problem.

PR #200, opened 2025-10-29
RFE-9629: Add copy to clipboard buttons to display token page

vojtechszocs:
This PR modifies the [token display page](https://github.com/openshift/oauth-server/blob/master/pkg/server/tokenrequest/tokenrequest.go) by adding "Show" and "Copy" buttons.

Initially, all snippets have their content hidden using `***` placeholder text.

<br>

Click on "Show" button to reveal the content and make the "Show" button disappear for that snippet.

<br>

Click on "Copy" button to copy the content to clipboard using the standard [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard).

<br>

Clicking the "Copy" button changes its text and disables it for 3 seconds, then resets the button to its normal state.
- success - shows "Copied!"
- error - shows "Error!" - note :memo: error is logged to browser console using `console.error`

## Summary by CodeRabbit

* **New Features**
  * Token and command displays now feature interactive "Copy" and "Show" buttons for improved usability.
  * Copy button provides immediate visual feedback confirming successful copying or indicating errors.
  * Command text rendering now generates on the client-side, enabling better interactivity and visual presentation.

### DIFF
--- a/pkg/server/tokenrequest/tokenrequest.go
+++ b/pkg/server/tokenrequest/tokenrequest.go
@@ -237,32 +237,36 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
   {{ .Error }}
 {{ else }}
   <script>
-    function codeSnippet(e, textBlocks) {
-      const snippet = textBlocks.join(' ');
+    function codeSnippet(e, textBlocks = []) {
+      const snippetString = textBlocks.join(' ').trim();
 
       const snippetHTML = textBlocks
-        .map((text) => '<span class="nowrap">' + text + '</span>')
+        .map((text) => {
+          const span = document.createElement('span');
+          span.className = 'nowrap';
+          span.innerText = text;
+          return span.outerHTML;
+        })
         .join(' ');
 
       const copyButton = document.createElement('button');
       copyButton.innerText = 'Copy';
 
+      const resetCopyButton = () => {
+        copyButton.innerText = 'Copy';
+        copyButton.disabled = false;
+      };
+
       copyButton.onclick = () => {
         copyButton.disabled = true;
-        navigator.clipboard.writeText(snippet)
+        navigator.clipboard.writeText(snippetString)
           .then(() => {
             copyButton.innerText = 'Copied!';
-            setTimeout(() => {
-              copyButton.innerText = 'Copy';
-              copyButton.disabled = false;
-            }, 3000);
+            setTimeout(resetCopyButton, 3000);
           })
           .catch((error) => {
             copyButton.innerText = 'Error!';
-            setTimeout(() => {
-              copyButton.innerText = 'Copy';
-              copyButton.disabled = false;
-            }, 3000);
+            setTimeout(resetCopyButton, 3000);
             console.error('Failed to copy snippet to clipboard', error);
           });
       };
--- a/pkg/server/tokenrequest/tokenrequest.go
+++ b/pkg/server/tokenrequest/tokenrequest.go
@@ -219,6 +219,8 @@ const cssStyle = `
 	code     { font-weight: 300; font-size: 1.5em; margin-bottom: 1em; display: inline-block;  color: #646464;  }
 	pre      { padding-left: 1em; border-radius: 5px; color: #003d6e; background-color: #EAEDF0; padding: 1.5em 0 1.5em 4.5em; white-space: normal; text-indent: -2em; }
 	pre>button { margin-left: 1.5em; margin-right: 1.5em; float: right; }
+	pre>button:disabled { color: #444; }
+	pre>button:disabled:hover { text-decoration: none; cursor: default; }
 	a        { color: #00f; text-decoration: none; }
 	a:hover  { text-decoration: underline; }
 	button   { background: none; border: none; color: #00f; text-decoration: none; font: inherit; padding: 0; }
@@ -234,31 +236,77 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
 {{ if .Error }}
   {{ .Error }}
 {{ else }}
-  <h2>Your API token is</h2>
-  <code>{{.AccessToken}}</code>
-
   <script>
-    var commands = {
-      login: 'oc login --token={{.AccessTokenJSStr}} --server={{.PublicMasterURLJSStr}}',
-      apiCall: 'curl -H "Authorization: Bearer {{.AccessTokenJSStr}}" "{{.PublicMasterURLJSStr}}/apis/user.openshift.io/v1/users/~"',
-    };
+    function codeSnippet(e, textBlocks) {
+      const snippet = textBlocks.join(' ');
+
+      const snippetHTML = textBlocks
+        .map((text) => '<span class="nowrap">' + text + '</span>')
+        .join(' ');
+
+      const copyButton = document.createElement('button');
+      copyButton.innerText = 'Copy';
+
+      copyButton.onclick = () => {
+        copyButton.disabled = true;
+        navigator.clipboard.writeText(snippet)
+          .then(() => {
+            copyButton.innerText = 'Copied!';
+            setTimeout(() => {
+              copyButton.innerText = 'Copy';
+              copyButton.disabled = false;
+            }, 3000);
+          })
+          .catch((error) => {
+            copyButton.innerText = 'Error!';
+            setTimeout(() => {
+              copyButton.innerText = 'Copy';
+              copyButton.disabled = false;
+            }, 3000);
+            console.error('Failed to copy snippet to clipboard', error);
+          });
+      };
+
+      const showButton = document.createElement('button');
+      showButton.innerText = 'Show';
+
+      showButton.onclick = () => {
+        e.innerHTML = snippetHTML;
+        e.appendChild(copyButton);
+      };
+
+      e.innerHTML = '***';
+      e.appendChild(copyButton);
+      e.appendChild(showButton);
+    }
   </script>
 
+  <h2>Your API token is</h2>
+  <pre id="token"></pre>
+
   <h2>Log in with this token</h2>
-  <pre>
-    oc login
-    <span class="nowrap">--token={{.AccessToken}}</span>
-    <span class="nowrap">--server={{.PublicMasterURL}}</span>
-    <button onclick="navigator.clipboard.writeText(commands.login)">Copy</button>
-  </pre>
+  <pre id="login"></pre>
 
   <h3>Use this token directly against the API</h3>
-  <pre>
-    curl
-    <span class="nowrap">-H "Authorization: Bearer {{.AccessToken}}"</span>
-    <span class="nowrap">"{{.PublicMasterURL}}/apis/user.openshift.io/v1/users/~"</span>
-    <button onclick="navigator.clipboard.writeText(commands.apiCall)">Copy</button>
-  </pre>
+  <pre id="apiCall"></pre>
+
+  <script>
+    codeSnippet(document.getElementById('token'), [
+      '{{.AccessTokenJSStr}}',
+    ]);
+
+    codeSnippet(document.getElementById('login'), [
+      'oc login',
+      '--token={{.AccessTokenJSStr}}',
+      '--server={{.PublicMasterURLJSStr}}',
+    ]);
+
+    codeSnippet(document.getElementById('apiCall'), [
+      'curl',
+      '-H "Authorization: Bearer {{.AccessTokenJSStr}}"',
+      '"{{.PublicMasterURLJSStr}}/apis/user.openshift.io/v1/users/~"',
+    ]);
+  </script>
 {{ end }}
 
 <br><br>
--- a/pkg/server/tokenrequest/tokenrequest.go
+++ b/pkg/server/tokenrequest/tokenrequest.go
@@ -144,6 +144,8 @@ func (t *tokenRequest) displayTokenPost(osinOAuthClient *osincli.Client, w http.
 	}
 
 	data.AccessToken = accessData.AccessToken
+	data.AccessTokenJSStr = template.JSStr(data.AccessToken)
+	data.PublicMasterURLJSStr = template.JSStr(data.PublicMasterURL)
 	renderToken(w, data)
 }
 
@@ -178,9 +180,11 @@ type sharedData struct {
 type tokenData struct {
 	sharedData
 
-	AccessToken     string
-	PublicMasterURL string
-	LogoutURL       string
+	AccessToken          string
+	AccessTokenJSStr     template.JSStr
+	PublicMasterURL      string
+	PublicMasterURLJSStr template.JSStr
+	LogoutURL            string
 }
 
 func getBaseURL(req *http.Request) (*url.URL, error) {
@@ -214,6 +218,7 @@ const cssStyle = `
 	code,pre { font-family: Menlo, Monaco, Consolas, monospace; }
 	code     { font-weight: 300; font-size: 1.5em; margin-bottom: 1em; display: inline-block;  color: #646464;  }
 	pre      { padding-left: 1em; border-radius: 5px; color: #003d6e; background-color: #EAEDF0; padding: 1.5em 0 1.5em 4.5em; white-space: normal; text-indent: -2em; }
+	pre>button { margin-left: 1.5em; margin-right: 1.5em; float: right; }
 	a        { color: #00f; text-decoration: none; }
 	a:hover  { text-decoration: underline; }
 	button   { background: none; border: none; color: #00f; text-decoration: none; font: inherit; padding: 0; }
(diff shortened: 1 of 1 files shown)


pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -233,10 +254,10 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
   <code>{{.AccessToken}}</code>
 
   <h2>Log in with this token</h2>
-  <pre>oc login <span class="nowrap">--token={{.AccessToken}}</span> <span class="nowrap">--server={{.PublicMasterURL}}</span></pre>
+  <pre>oc login <span class="nowrap">--token={{.AccessToken}}</span> <span class="nowrap">--server={{.PublicMasterURL}}</span><button role="copy-to-clipboard">Copy</button></pre>
```
spadgett: I don't think `copy-to-clipboard` is a valid role.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -224,6 +225,26 @@ const cssStyle = `
...
+<script>
+  const isCopyButton = (node) => node.localName === "button"
+    && node.attributes.getNamedItem("role") !== null
+    && node.attributes.getNamedItem("role").value === "copy-to-clipboard";
+
+  document.querySelectorAll("button[role=copy-to-clipboard]").forEach((button) => {
+    button.onclick = () => {
+      const textToCopy = Array.from(button.parentElement.childNodes.values()).reduce((acc, node) => {
+        return isCopyButton(node) ? acc : acc + node.textContent;
+      }, "");
```
spadgett: This works, but I wonder if it's not simpler to do something like

```
navigator.clipboard.writeText('oc login --token={{.AccessToken}} --server={{.PublicMasterURL}}');
```

in the button click handler. It repeats the command, but seems less error-prone.

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -224,6 +225,26 @@ const cssStyle = `
...
+
+      if (navigator.clipboard) {
```
spadgett: nitpick: I'd probably remove this check. Otherwise clicking the button will just result in nothing happening and no error when unsupported. If we do keep the check, it should be to hide the button. But since the feature is supported in all modern browsers, I'd skip it.

https://caniuse.com/?search=navigator.clipboard

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: This custom role was meant as a semantic extension of the standard `button` role, in order to identify any "Copy" button instances in the document. I thought that using custom HTML `role` values is not a bad thing, assuming they don't conflict with standard `role` values from the HTML spec (?)

To my understanding, `class` is for handling visuals and `role` is for handling semantics when working with DOM elements. Is there some alternative way to identify "Copy" button instances you'd suggest in this particular case?

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: Fair point, I'll update the code and try to reduce the accidental complexity in JS code.

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: OK, I will remove the check.

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: As for the reasoning behind the code - I wanted to keep it simple - the given `<pre>` element contains the snippet we want to be copied, so we add `<button>` right into it and the button takes the content of the `<pre>` element (minus the "Copy" button text) and copies it to clipboard.

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: Update - we don't need the `role` attribute anymore.

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: Code updated to directly call `navigator.clipboard.writeText` as suggested.

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -232,11 +233,28 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
   <h2>Your API token is</h2>
   <code>{{.AccessToken}}</code>
 
+  <script>
+    var cmd = {
+      login: 'oc login --token={{.AccessToken}} --server={{.PublicMasterURL}}',
+      api: 'curl -H "Authorization: Bearer {{.AccessToken}}" "{{.PublicMasterURL}}/apis/user.openshift.io/v1/users/~"',
```
spadgett: @everettraven Do you know if there is ever a danger of the character `'` appearing in `AccessToken` or `PublicMasterURL`? Is there any way to escape it?

vojtechszocs: If needed, we can preemptively add code to replace `'` with `\'` within the command strings.

everettraven: As long as the access token is a JWT, I don't believe so. AFAIK, the only characters present in a JWT would be `a-z`, `A-Z`, `0-9`, `-`, `.`, and `_`.

I don't suspect we would end up with a non-JWT token but I don't have as much historical context here as @ibihim would.

spadgett: It looks like we can use `JSStr` to handle any special characters?

https://pkg.go.dev/html/template#JSStr

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -233,10 +250,16 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
   <code>{{.AccessToken}}</code>
 
   <h2>Log in with this token</h2>
-  <pre>oc login <span class="nowrap">--token={{.AccessToken}}</span> <span class="nowrap">--server={{.PublicMasterURL}}</span></pre>
+  <pre>
+    {{range $p := .LoginParts}}<span class="nowrap">{{$p}}</span>{{end}}
```
spadgett: Will this break escaping of HTML since it's now a `JSStr`?

I was thinking of just having `AccessToken` and `AccessTokenJSStr` in the template data and using the first in the HTML and the second in the JS copy clipboard code. FWIW the duplication doesn't bother me since one command has HTML markup and the other doesn't. They're not exactly the same. It's almost better to be explicit.

vojtechszocs: > Will this break escaping of HTML since it's now a `JSStr`?

Hm, you are right, the parser will detect `<span class="nowrap">{{ something }}</span>` and will try to HTML escape `something` automatically which is probably not what we want.

OK, I will add simple strings instead of using "parts" as suggested.

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -232,11 +235,28 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
   <h2>Your API token is</h2>
   <code>{{.AccessToken}}</code>
 
+  <script>
+    var commands = {
+      login: 'oc login --token={{.AccessTokenJSStr}} --server={{.PublicMasterURL}}',
+      apiCall: 'curl -H "Authorization: Bearer {{.AccessTokenJSStr}}" "{{.PublicMasterURL}}/apis/user.openshift.io/v1/users/~"',
```
spadgett: I'd suggest we have a `PublicMasterURLJSStr` as well

spadgett (approved):
/lgtm

Thanks @vojtechszocs 👍 

/hold
for @everettraven to review

everettraven (approved):
This seems fine to me. I don't think I have approval permissions. Will likely need @ibihim for approval.

/lgtm

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -232,11 +237,28 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
   <h2>Your API token is</h2>
   <code>{{.AccessToken}}</code>
 
+  <script>
+    var commands = {
```
ibihim: Maybe this is my old school JS, but isn't it better to have something like this namespaced approach:

```js
  var openshift = openshift || {};
  openshift.commands = openshift.commands || {};
  openshift.commands.login = '...';
  openshift.commands.apiCall = '...';
```

or a IIFE that is completely isolated?

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -232,11 +237,28 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
...
+    };
+  </script>
+
   <h2>Log in with this token</h2>
-  <pre>oc login <span class="nowrap">--token={{.AccessToken}}</span> <span class="nowrap">--server={{.PublicMasterURL}}</span></pre>
+  <pre>
+    oc login
+    <span class="nowrap">--token={{.AccessToken}}</span>
+    <span class="nowrap">--server={{.PublicMasterURL}}</span>
+    <button onclick="navigator.clipboard.writeText(commands.login)">Copy</button>
```
ibihim: Shouldn't we add some feedback for the user, like changing the button text to "Copied!" or so?

ibihim: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

Isn't it best practice to handle the error somehow?

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: Before this PR, the token request page had no JavaScript so I'm not sure if using the `openshift` variable to namespace the commands is really needed.

I wanted to keep the changes at minimum so I went with a single `commands` variable to hold the strings to copy.

```js
var commands = { foo: 'x' };
```
is functionally the same as
```js
var commands = {};
commands.foo = 'x';
```

As for IIFE, these are typically used to avoid pollution of the global object namespace but I don't think we need them in our case.

The idea is to define some global object to store the strings to be referenced by the copy button's `onclick` handler.

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: This would add more complexity, which is something I wanted to avoid.

The page template is missing standard HTML structure, including `<!doctype html>`, `<head>`, `<body>` etc. which cause browsers to render this page it in "quirks" mode.

I could try to add some custom-styled "Copied!" feedback text upon clicking the button, but do we really need it?

vojtechszocs: We just call the `clipboard.writeText` function without handling the resulting Promise.

This function is async and resolves when the clipboard's contents have been updated. If we had a "Copied!" feedback text implemented, we'd be handing the Promise, but currently we don't need that.

I'm not sure we need an explicit error handler here, unless we want to implement the feedback text, could be something like "Failed to copy the snippet to clipboard, see browser logs for details" or similar.

In practice, `clipboard.writeText` function could fail if the page was forbidden to copy text to clipboard via setting page permissions, e.g. for Chromium based browsers [1]:
> Writing requires either the `clipboard-write` permission or transient activation.

[1] https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations

spadgett: @vojtechszocs Maybe just throw up an alert on an error? Something like...? "Copy to clipboard failed. Select the content to copy it manually."

In practice, I'd expect clipboard write to almost never fail.

pkg/server/tokenrequest/tokenrequest.go
ibihim: I still think that it is always worth to introduce best practices, but nevertheless, this isn't a thing that I would keep blocking the PR.

pkg/server/tokenrequest/tokenrequest.go
ibihim: Implemented.

ibihim: Implemented

ibihim:
I wish there would have been a more detailed explanation why it is fine to get rid of the intermediate CSRF step, but after bouncing my head for an hour on this topic, the worst thing I could imagine is that a disgruntled Alice could sneak in her tokens on your page.

ibihim:
I will approve once @everettraven lgtm'ed it... or @liouk

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -76,41 +72,8 @@ func (t *tokenRequest) requestToken(osinOAuthClient *osincli.Client, w http.Resp
...
 
 func (t *tokenRequest) displayToken(osinOAuthClient *osincli.Client, w http.ResponseWriter, req *http.Request) {
-	switch req.Method {
-	case http.MethodGet:
-		t.displayTokenGet(osinOAuthClient, w, req)
-	case http.MethodPost:
-		t.displayTokenPost(osinOAuthClient, w, req)
-	default:
+	if req.Method != http.MethodGet {
 		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
```
everettraven, line 76: This seems like a breaking change - is there a particular reason we are making this change?

everettraven (commented):
I'm concerned about us making a breaking change to the acceptable requests here - because it is out there, and has been out there for a while, I would not be surprised if there is some kind of end user workflow that depends on sending a properly formed POST request to display the token.

Additionally, do we have any kind of testing in place for UX/UI related changes we can update / put in place here?

If there is not existing testing infrastructure here it isn't the end of the world, but if it exists we should make sure the UX changes are appropriately tested.

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -253,23 +263,6 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
...
-{{ if .Error }}
-  {{ .Error }}
-  <br><br>
-  <a href="{{.RequestURL}}">Request another token</a>
-{{ else }}
-  <form method="post" action="{{.Action}}">
-    <input type="hidden" name="code" value="{{.Code}}">
-    <input type="hidden" name="csrf" value="{{.CSRF}}">
-    <button type="submit">
-      Display Token
```
spadgett, line 267: We can't remove this. This was there to prevent cross-site request forgery.

/hold

spadgett (changes requested):

vojtechszocs:
Sorry, this PR got stuck in a limbo state.

@ibihim 

> I wish there would have been a more detailed explanation why it is fine to get rid of the intermediate CSRF step, but after bouncing my head for an hour on this topic, the worst thing I could imagine is that a disgruntled Alice could sneak in her tokens on your page.

AFAIK we have customer RFEs that suggested removing the intermediate "Display Token" form submit (CSRF) step since it's one more page to go through when accessing token data.

That said, I've removed this change from the PR to preserve the existing behavior.

@everettraven 

> I'm concerned about us making a breaking change to the acceptable requests here - because it is out there, and has been out there for a while, I would not be surprised if there is some kind of end user workflow that depends on sending a properly formed POST request to display the token.

That makes perfect sense :+1:

pkg/server/tokenrequest/tokenrequest.go
```diff
@@ -229,14 +236,77 @@ var tokenTemplate = template.Must(template.New("tokenTemplate").Parse(
 {{ if .Error }}
   {{ .Error }}
 {{ else }}
+  <script>
+    function codeSnippet(e, textBlocks) {
+      const snippet = textBlocks.join(' ');
+
+      const snippetHTML = textBlocks
+        .map((text) => '<span class="nowrap">' + text + '</span>')
```
spadgett: Should we encoding as HTML here to avoid potential XSS? Do the text blocks themselves ever have markup? If we never pass markup, it makes sense to me to always encode just to be safe.

Apologies if I asked this before, it's been a little while since the last review :)

everettraven (approved):
Changes seem fine to me. Will defer to @spadgett for any web-specific concerns.

pkg/server/tokenrequest/tokenrequest.go
vojtechszocs: @spadgett I think this depends on template data used to interpolate `{{ exp }}` expressions.

```ts
codeSnippet(document.getElementById('login'), [
    'oc login',
    '--token={{.AccessTokenJSStr}}',
    '--server={{.PublicMasterURLJSStr}}',
  ]);
```

In any case, I agree that it's best to stay safe and ensure that snippet content is treated as text only.

vojtechszocs:
PR updated to ensure that snippet content is treated as text only.

This can be tested by manually changing the snippet text blocks inside `tokenTemplate` for example:
```js
codeSnippet(document.getElementById('token'), [
  '{{.AccessTokenJSStr}} <foo>&test</foo>',
]);
```

pkg/server/tokenrequest/tokenrequest.go
spadgett: We should make sure we're not double-encoding the strings if the go template is doing it already.

vojtechszocs: @spadgett That should not be an issue since we're using `element.innerText` which treats whatever value is passed in as text only, i.e. there is no additional encoding or escaping involved.

spadgett (approved):
/lgtm

everettraven (approved):
/approve
/lgtm

vojtechszocs:
Changes in `tokenrequest.go` are client-side i.e. `<script>` code only - verified based on local browser testing.

/verified by @vojtechszocs

PR #201, opened 2025-11-18
OCPBUGS-64657: external/openid: Unwrap groups encoded as string

tchap:
Unwrap groups array claim when encoded as a simple string.
This means decoding `"groups": "[\"A\", \"B\"]"` into `["A", "B"]`
groups claim value.

I also made the array parsing more strict, failing on unexpected type encountered instead of ignore/panic.

Let me know if we need to add more tests somewhere.


ibihim:
/hold

This is a regression and has potential to backfire massively with parsing discrepancies.
There must be some more research what is going on.

In its current state, It could pass as a feature request hidden behind a config.

tchap:
I am going to close this as it needs further planning.

PR #205, opened 2025-12-11
WIP: OCPBUGS-63228: groupmapper: make direct API calls to retrieve groups instead of using a cache

liouk:
Per the [Synchronizing User Groups on Login](https://github.com/openshift/enhancements/blob/master/enhancements/authentication/synchronizing-user-groups-on-login.md) enhancement, OIDC groups are synchronized to Openshift groups upon user login. During synchronization, the oauth-server checks current user groups and compares with the ones obtained from the provider via the token; it then calculates which groups need to be added/removed to/from the user, and performs the changes accordingly.

There is currently an issue with this implementation, as the server uses a groups cache to determine user group membership; this cache can be stale, which can lead to problems such as [OCPBUGS-63228](https://issues.redhat.com//browse/OCPBUGS-63228).

During authorization, correctness is non-negotiable, so we must never rely on a cache to determine user group membership. This PR removes the group cache and replaces it with direct API calls; correctness comes at the cost of increased performance, especially in large clusters with large numbers of groups.

### DIFF
--- a/pkg/groupmapper/groupmapper.go
+++ b/pkg/groupmapper/groupmapper.go
@@ -3,19 +3,16 @@ package groupmapper
 import (
 	"context"
 	"fmt"
-	"time"
+	"slices"
 
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


liouk:
Closing in favor of #211.

/close

PR #211, opened 2026-02-25
OCPBUGS-63228: groupmapper: avoid mutating original group users slice when removing user

liouk:
When removing a user from a group, `removeUserFromGroup` gets a pointer to the group object living in the informer cache via the lister, then builds the new user list using append on a sub-slice of it:
```go
  newUsers = append(updatedGroup.Users[0:userIdx], updatedGroup.Users[userIdx+1:]...)
```

Because the sub-slice shares the backing array with the cached object, `append` mutates the cached object's user list in place -- the removed user disappears and the last user gets duplicated. When the informer re-indexes the group after an update event, it diffs the corrupted old object against the new one. Since the removed user is no longer in the corrupted old object, its index entry is never cleaned up and becomes a permanent phantom.

On subsequent logins, `GroupsFor(user)` returns the phantom group from the index. `groupsDiff` sees that the cache already includes the group, matches the provider's desired state, and computes no changes needed -- so the user is never actually added to the group.

`addUserToGroup` has a similar issue. 

This PR fixes this issue by creating a slice with a new backing array to store the updated user list. `addUserToGroup` can simply use the deep-copied object as this already uses a new backing array.

### DIFF
--- a/pkg/groupmapper/groupmapper.go
+++ b/pkg/groupmapper/groupmapper.go
@@ -3,6 +3,7 @@ package groupmapper
 import (
 	"context"
 	"fmt"
+	"slices"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


liouk:
Since we do not currently have any unit tests that cover this issue, I will create a proof PR to demonstrate the bug and then incorporate the appropriate unit tests in this one. 

/hold

liouk:
Proof PR: #212 
Failing unit tests: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oauth-server/212/pull-ci-openshift-oauth-server-master-unit/2026678658705395712

These tests have also been added in this PR, and must succeed for the fix to be proven.

liouk:
Unit tests on proof PR (#212) failed as expected, while the ones in this PR are passing.

/hold cancel

everettraven (approved):
/lgtm

liouk:
Verified via:
- targeted unit tests in this PR and the proof PR (#212); discussed [above](https://github.com/openshift/oauth-server/pull/211#issuecomment-3960114676)
- I've also been reproducing the issue using [this script](https://github.com/liouk/redhat/blob/main/issues/OCPBUGS-63228_oauth-oidc-groups-lost/reproduce.sh) -- eventually after ~100 iterations we hit this problem. With the fix in place, I've never been able to reproduce this, even after running the script for hours (2k+ iterations)

Marking as verified based on the above.

/verified by @liouk

cc @xingxingxia

PR #215, opened 2026-03-08
CNTRLPLANE-2446: K8s 1.35 rebase

ingvagabund:
Evidence for https://github.com/openshift/kubernetes-apiserver/pull/84

Picking up https://github.com/openshift/oauth-server/pull/207

## Summary by CodeRabbit

* **Tests**
  * Updated group-mapper tests to better mirror Kubernetes 1.35+ informer behavior, including correct handling of initial watch bookmark events and cache synchronization.
  * Improved test setup to avoid timing/order issues between bookmark processing and initial list results.
* **Chores**
  * Updated the pinned Kubernetes/API server dependency version.


ingvagabund:
ci/prow/e2e-gcp failing with:
```
[sig-auth][Feature:SCC][Early] should not have pod creation failures during install [Suite:openshift/conformance/parallel]
[sig-network] LoadBalancers [Feature:LoadBalancer] should be able to preserve UDP traffic when server pod cycles for a LoadBalancer service on different nodes
```
previously with:
```
[sig-auth][Feature:OAuthServer] [Token Expiration] Using a OAuth client with a non-default token max age [apigroup:oauth.openshift.io] to generate tokens that do not expire works as expected when using a code authorization flow [apigroup:user.openshift.io] [Suite:openshift/conformance/parallel]
```
after that:
```
[sig-auth][Feature:RoleBindingRestrictions] RoleBindingRestrictions should be functional Create a RBAC rolebinding when subject is not already bound and is not permitted by any RBR should fail [apigroup:authorization.openshift.io] [Suite:openshift/conformance/parallel]
```

PR #244, opened 2026-06-30
CNTRLPLANE-3751: Add dynamic proxy CA reload for outbound IdP transports

tchap:
The proxy CA path is read from the ProxyTrustedCA field on the OAuth
config struct. When set, the OAuth Server watches the proxy CA file on
disk and rebuilds outbound HTTP transports when it changes. The
transport's RootCAs starts from the system cert pool and appends any
static IdP CA and the dynamic proxy CA, so public-CA IdPs remain
trusted alongside TLS-intercepting or HTTPS proxies.

Static TLS material (IdP CA certs, client certificates) is loaded once
at construction time; only the proxy CA is reloaded dynamically. On
reload failure the old transport is preserved.

## Summary by CodeRabbit

* **New Features**
  * Added dynamic proxy CA trust updates for OAuth and password-based sign-in. Trust can be configured via `PROXY_TRUSTED_CA_FILE` and applied after startup without restarting.
* **Bug Fixes**
  * Improved TLS transport handling with safer defaults, stricter validation for certificate/key pairing, and continued operation when dynamic proxy CA updates are invalid.
* **Tests**
  * Expanded coverage for dynamic CA trust selection, combined CA trust, reload behavior, error resilience, and default/negative cases.


pkg/oauthserver/dynamic_transport_test.go
```diff
@@ -0,0 +1,266 @@
...
+	proxyCAFile := filepath.Join(t.TempDir(), "proxy-ca.pem")
+	writeFile(t, proxyCAFile, proxyCAPEM)
+
+	rt, err := newDynamicCARoundTripper(proxyCAFile, "", "", "")
+	if err != nil {
+		t.Fatalf("newDynamicCARoundTripper: %v", err)
+	}
+
+	req, _ := http.NewRequest("GET", server.URL, nil)
+	resp, err := rt.RoundTrip(req)
```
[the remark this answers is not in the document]
tchap: This is now fixed.

pkg/oauthserver/dynamic_transport.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	proxyCABundle := rt.proxyCAContent.CurrentCABundleContent()
+	if !roots.AppendCertsFromPEM(proxyCABundle) {
+		return nil, fmt.Errorf("failed to parse proxy CA bundle")
+	}
+
+	t := knet.SetTransportDefaults(&http.Transport{
+		TLSClientConfig: &tls.Config{RootCAs: roots},
+	})
```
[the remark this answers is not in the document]
tchap: This is now fixed.

pkg/oauthserver/dynamic_transport_test.go
```diff
@@ -0,0 +1,323 @@
...
+	if err == nil {
+		t.Fatal("expected RoundTrip to server2 to fail with proxyCA1")
+	}
+
+	// Rotate proxy CA to CA2, let the fsnotify watcher pick it up
+	writeFile(t, proxyCAFile, proxyCA2PEM)
+
+	err = wait.PollUntilContextTimeout(t.Context(), 50*time.Millisecond, 10*time.Second, false, func(ctx context.Context) (bool, error) {
+		_, err := doRequest(t, rt, server2.URL)
+		return err == nil, nil
```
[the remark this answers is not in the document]
tchap: Fixed for resp.Body. Ignoring Close errors is normal, skipping that.

pkg/oauthserver/dynamic_transport_test.go
```diff
@@ -0,0 +1,328 @@
...
+	proxyCAFile := filepath.Join(dir, "proxy-ca.pem")
+	writeFile(t, idpCAFile, idpCAPEM)
+	writeFile(t, proxyCAFile, proxyCAPEM)
+
+	rt, err := newDynamicCARoundTripper(proxyCAFile, idpCAFile, "", "")
+	if err != nil {
+		t.Fatalf("newDynamicCARoundTripper: %v", err)
+	}
+
+	mustGet(t, rt, idpServer.URL)
```
ehearne-redhat: I'm a little unsure about the usage of `mustGet()` because while it is valuable to not repeat code, using it generically will result in errors that are not as useful. 

What I would find more useful is perhaps getting the helper function to have return type `err` so we can detect something went wrong and report it clearly in tests.

pkg/oauthserver/dynamic_transport_test.go
```diff
@@ -0,0 +1,328 @@
...
+	t.Helper()
+	resp, err := doRequest(t, rt, url)
+	if err != nil {
+		t.Fatalf("GET %s: %v", url, err)
+	}
+	_ = resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		t.Fatalf("GET %s: expected 200, got %d", url, resp.StatusCode)
+	}
+}
```
ehearne-redhat:
```suggestion
func mustGet(t *testing.T, rt http.RoundTripper, url string) (error) {
	t.Helper()
	resp, err := doRequest(t, rt, url)
	if err != nil {
		return err
	}
	_ = resp.Body.Close()
	if resp.StatusCode != http.StatusOK {
		return fmt.Errorf("GET %s: expected 200, got %d", url, resp.StatusCode)
	}
}
```

pkg/oauthserver/dynamic_transport_test.go
```diff
@@ -0,0 +1,328 @@
...
+	writeFile(t, idpCAFile, idpCAPEM)
+	writeFile(t, proxyCAFile, proxyCAPEM)
+
+	rt, err := newDynamicCARoundTripper(proxyCAFile, idpCAFile, "", "")
+	if err != nil {
+		t.Fatalf("newDynamicCARoundTripper: %v", err)
+	}
+
+	mustGet(t, rt, idpServer.URL)
+	mustGet(t, rt, proxyServer.URL)
```
ehearne-redhat: Example:

In some cases, it might not be necessary to error out. Feel free to make the judgement call there. :) 

```suggestion
	err = mustGet(t, rt, proxyServer.URL)
	t.Fatalf("error performing GET on proxy %v: %v", proxyServer.URL, err)
```

ehearne-redhat (commented):
My main concern is with `mustGet()` . Also, while not completely necessary, what do you think of adding a test in oauth_apiserver_test.go that:
  1. Creates an OAuthServerConfig with PROXY_TRUSTED_CA_FILE set
  2. Calls methods that trigger transportForInner (this happens during WithOAuth)
  3. Verifies the post-start hook exists and can be called without error
  4. Confirms dynamicTransports contains the expected entries

I wouldn't block on the PR if you didn't think it was required. Just wanted to know your thoughts :)

tchap:
> My main concern is with `mustGet()` . Also, while not completely necessary, what do you think of adding a test in oauth_apiserver_test.go that:
> 
> 1. Creates an OAuthServerConfig with PROXY_TRUSTED_CA_FILE set
> 2. Calls methods that trigger transportForInner (this happens during WithOAuth)
> 3. Verifies the post-start hook exists and can be called without error
> 4. Confirms dynamicTransports contains the expected entries
> 
> I wouldn't block on the PR if you didn't think it was required. Just wanted to know your thoughts :)

I reorganized things a little bit. Updated `TestTransportForInner_WithProxyCA` to really test things as completely as possible, including the env variable, given the fact that the code does not provide very good interface for hooking things up together.

Also extended `mustGet` to accept a context message, which I don't think is very necessary as you can always find the file line, but ok, no big deal adding it. So I don't think there is any other context missing and I will generally keep `mustGet` usage as is.

everettraven (commented):
Aside from the relatively minor comment on using an environment variable instead of writing the proxy CA file path to the config, this seems pretty straightforward.

pkg/oauthserver/oauth_apiserver.go
```diff
@@ -282,6 +291,23 @@ type ExtraOAuthConfig struct {
...
+	proxyTrustedCA string
+	// dynamicTransports collects transports that watch the proxy CA file
+	// for hot-reload. Populated by transportForInner during handler init,
+	// then started via the StartDynamicCAWatchers post-start hook.
+	dynamicTransports []*dynamicCARoundTripper
+}
+
+func (e *ExtraOAuthConfig) initProxyCA() {
+	e.proxyTrustedCA = os.Getenv("PROXY_TRUSTED_CA_FILE")
+}
```
everettraven: Instead of using an environment variable, should this instead be a configuration option added to the configuration file API surface?

tchap: I am working on this. Updated openshift/api to contain ProxyTrustedCA, now I need to rebase onto 1.35 to be able to update openshift/api. Tracked in https://github.com/openshift/oauth-server/pull/247

tchap: Implemented.

ehearne-redhat (commented):
/lgtm

pkg/oauthserver/auth.go
```diff
@@ -762,25 +762,40 @@ func (redirectSuccessHandler) AuthenticationSucceeded(user kuser.Info, then stri
...
 		return nil, errors.New("certFile and keyFile must be specified together")
 	}
 
-	// Copy default transport
+	if len(proxyCA) != 0 {
+		rt, err := newDynamicCARoundTripper(proxyCA, ca, certFile, keyFile)
+		if err != nil {
+			return nil, err
+		}
+		c.ExtraOAuthConfig.dynamicTransports = append(c.ExtraOAuthConfig.dynamicTransports, rt)
```
ibihim: We need to do some rework here.

```go
proxyCA := c.ExtraOAuthConfig.Options.ProxyTrustedCA
```

Law of demeter: I think this should be a bad smell, if you reach 3 levels deep. When this object changes, we need to change as well. Handing the "ProxyTrustedCA" would have been better...

... but there is more going on here, we have now a side-effect. Before we had a simple function and now we hook up dynamic content. This is not so good. Ideally we see from the function signature what is going on.
**Also we have 6 dynamic file listeners on the same file.**

I think it is strictly better to create the dynamic file listener once in `getOAuthProvider`, append it there to the dynamic config thingy and then hand it down to the dynamicCARoundTripper for re-use.

tchap: I will check out the listeners, I am sure I noticed this issue but for some reason didn't pursue to improve it.

`ProxyTrustedCA` is fine in general, I think, it can only change with server restarts.

ibihim: Correction to my proposal below: getOAuthProvider isn't "once". It runs per IdP. So maybe somewhere else.

tchap: Reimplemented in b659db4 . The main part really is in `configureTransport`.

pkg/oauthserver/oauth_apiserver.go
```diff
@@ -282,6 +288,32 @@ type ExtraOAuthConfig struct {
...
 
 	postStartHooks map[string]genericapiserver.PostStartHookFunc
+
+	// transportBuilderFunc creates an http.RoundTripper for outbound IdP requests.
+	// Set at init time by configureTransport: when a proxy CA is configured, it produces
+	// dynamicCARoundTrippers that share a single file watcher; otherwise it produces static transports.
+	transportBuilderFunc func(ca, certFile, keyFile string) (http.RoundTripper, error)
+}
+
+func configureTransport(config *ExtraOAuthConfig, proxyCAFile string) error {
```
everettraven, line 298: Should we have a nil check here for `config`?

everettraven (commented):
Aside from a minor question, this LGTM

pkg/oauthserver/oauth_apiserver.go
tchap: It can never be `nil` in production code...

PR #253, opened 2026-09-01
OCPBUGS-115307: bugfix: default to english when Accept-Language header contains more than 1000 underscores

everettraven:
to prevent a quadratic-time denial-of-service attack from spending a significant amount of time attempting to parse the acceptable languages from the request header.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed language preference headers, including mixed hyphen and underscore formatting.
  * Requests with excessively complex language formatting now safely fall back to English.
  * Added safeguards for headers containing more than 1,000 combined separators, preventing parsing issues and ensuring a consistent default language.

### DIFF
--- a/pkg/server/locales/locales.go
+++ b/pkg/server/locales/locales.go
@@ -1,6 +1,9 @@
 package locales
 
 import (
+	"errors"
+	"strings"
+
 	"golang.org/x/text/language"
 	"k8s.io/klog/v2"
 )
@@ -25,6 +28,17 @@ func GetLocale(acceptLangHeader string) Localization {
 }
 
 func getPreferredLang(acceptLangHeader string) string {
+	// OCPBUGS-92015: In order to prevent the Quadratic-time DoS attack
+	// that is possible as documented by https://github.com/golang/go/issues/79684,
+	// return early with the fallback to english if there are more than 1000 underscore and hyphen
+	// characters in the Accept-Language header.
+	// This can be removed once we have updated the language dependency to a version
+	// that has fixed this issue.
+	if strings.Count(acceptLangHeader, "-")+strings.Count(acceptLangHeader, "_") > 1000 {
+		klog.V(5).Infof("Error parsing 'Accept-Language' header, falling back to English language: %v", errors.New("tag list exceeds max length"))
+		return language.English.String()
+	}
+
 	matcher := language.NewMatcher(supportedLangs)
 	userPrefs, _, err := language.ParseAcceptLanguage(acceptLangHeader)
 	if err != nil {
--- a/pkg/server/locales/locales_test.go
+++ b/pkg/server/locales/locales_test.go
@@ -2,6 +2,7 @@ package locales
 
 import (
 	"reflect"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/server/locales/locales.go
```diff
@@ -25,6 +28,17 @@ func GetLocale(acceptLangHeader string) Localization {
 }
 
 func getPreferredLang(acceptLangHeader string) string {
+	// OCPBUGS-92015: In order to prevent the Quadratic-time DoS attack
+	// that is possible as documented by https://github.com/golang/go/issues/79684,
+	// return early with the fallback to english if there are more than 1000 underscore
+	// characters in the Accept-Language header.
+	// This can be removed once we have updated the language dependency to a version
+	// that has fixed this issue.
+	if strings.Count(acceptLangHeader, "_") > 1000 {
```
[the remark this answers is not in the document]
liouk: AFAIU, `ParseAcceptLanguage` already counts hyphens.

pkg/server/locales/locales_test.go
```diff
@@ -51,6 +52,11 @@ func TestLocales(t *testing.T) {
 			header: "cz;q=0.5, de;q=0.8",
 			locale: locale_en,
 		},
+		{
+			name: "Test 'Accept-Language' request header with too many underscores, so defaults to English language",
+			header: strings.Repeat("_", 2000),
+			locale: locale_en,
+		},
```
liouk, line 59: This test passes even without the new guard; `ParseAcceptLanguage()` would return an error because the value isn't a valid language. I understand that the test now passes because of the new guard, but I'm worried about regressions if e.g. we break the guard accidentally.

However, to be honest, I don't know whether it's possible to write a test that will help prevent such regressions, because the function's output doesn't help distinguish the cases (unless we capture and check the logged error somehow).

Therefore I won't block the PR on this.

liouk (commented):
@everettraven I'll let you decide on the two outstanding issues (my comment below, and the [one from @coderabbitai](https://github.com/openshift/oauth-server/pull/253#discussion_r3912082236), I won't block the PR on either.

/lgtm
/hold

pkg/server/locales/locales.go
ibihim: Why do we count up to a 1000  "_"? I know we copy-pasted it from a PR on upstream Go, but still makes me wonder. In our case more than 2 would be odd, wouldn't it be? Lets say then count till 10... but 1000?

pkg/server/locales/locales_test.go
everettraven: Yeah, without a refactor here such that the function actually returns an error case I'm not sure there is a way to reliably prove that we are testing the exact path we are intending other than making sure we are getting our fallback language.

In order to _properly_ test this here, we would have to construct a language header with >1000 language tags separated by `_` and I'm not sure that is actually possible without significantly more work (I don't think go is aware of that many language tags).

I'm not sure the additional verification is worth the squeeze here but :shrug:

pkg/server/locales/locales.go
everettraven: I'm following the same reasoning as the original fix for this issue that added a cap to the number of hyphens - https://github.com/golang/text/commit/434eadcdbc3b0256971992e8c70027278364c72c

TLDR; My understanding is that it is long enough for common use cases with significant buffer room to ideally not break potential edge cases that need more than a handful of tags. It seems like the cap of 1000 successfully mitigates the DoS attack vector while limiting the potential for breaking uses of it.

everettraven: We don't necessarily have the same concerns, but I'd rather be consistent with what upstream Go has decided so we can just adopt whatever fix goes into the package later.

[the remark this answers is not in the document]
everettraven: Good catch coderabbit - I had missed this path. Should be fixed now.

ibihim (commented):
/approve

